PR #12, opened 2014-09-16
[hydro] add 2dof test rig urdf

floweisshardt:


fmessmer:
LGTM, but we develop ros_canopen only against indigo...

PR #44, opened 2014-12-05
Halt works again

mathias-luedtke:
Fixed init ordering bug, added Halt to 402 again.


mathias-luedtke:
currenlty does not work properly with Elmo config, do not merge

floweisshardt:
visual review: OK

PR #66, opened 2015-02-11
DCF overlay feature

mathias-luedtke:
Overlay DCF parameters with settings from ROS parameter server, example given in canopen_test_utils/config/canopen_rig1.yaml
Needed for https://github.com/ipa320/cob_robots/pull/269

PR #65, opened 2015-02-11
Elmo console tool + bug fixes

mathias-luedtke:
- EMCY handling depends on entry availibity
- fixed SDO handling, especially for strings
- some clean-up
- added canopen_elmo_console: stream in Elmo commands and get responses

Issue #67, opened 2015-02-12
difference between ELMO and SCHUNK


floweisshardt:
in canopen_master/src/node.cpp we currently need

for SCHUNK:

```
        reset();
```

and for ELMO (if more than one module is on the bus):

```
        reset_com();
        emcy_.recover();
```

@thiagodefreitas, @ipa-mdl: please think about a solution which works for both

thiagodefreitas:
For running the test on the cob4-6, I had to do the following changes:
- canopen_master/src/node.cpp
  void Node::init(LayerStatus &status){
     nmt_listener_ = interface_->createMsgListener( can::MsgHeader(0x700 + node_id_), can::CommInterface::FrameDelegate(this, &Node::handleNMT));
  
     sdo_.init();
     try{
        reset_com();
        emcy_.recover();
         prepare();
     } 
     ...
- canopen_master/src/emcy.cpp
  
    void EMCYHandler::read(LayerStatus &status)
    {
    /*
         if(error_register() & (~32)){ // mask device specific error
             status.error("Node has emergency error");
         }
        */

mathias-luedtke:
This disables the error detection (incl. emergency stop) completely!
The EMCY handing and especially the reset need some more work, I will try to find some time for it next week.

floweisshardt:
current state should work with both. needs to be tested

mathias-luedtke:
Current implementation should work for most devices, if not specific issues should be filed.

PR #69, opened 2015-02-16
updates from cob4-2

floweisshardt:
includes #65 and #66. 
tested on cob4-2 arms: OK


floweisshardt:
I'll accept this PR as it was tested on cob4-2. to work on Elmo again, #67 still needs to be solved.

PR #72, opened 2015-03-02
Removed SimpleLayer

mathias-luedtke:
SimpleLayer did not provide logging, was only created for code transition.


mathias-luedtke:
tested on cob4-6: OK

Issue #74, opened 2015-03-09
Add support for hex numbers in eds file


mistoll:
Some vendors use hex numbers for "Unsigned#" data types in eds files. It would greatly enhance the compatibliity if the parser would accept these values. 
And according to CanOpen 306 chapter 4.3 this should be supported.

Example EDS:
  [MandatoryObjects]
  SupportedObjects=0x03
  1=0x1000
  2=0x1001
  3=0x1018

For the whole file see: http://en.nanotec.com/products/1604-pd4-c-high-pole-plug-drive-dc-servo-motor/

mathias-luedtke:
Thanks for the hint!

I have implemented a fix that should read all numbers with hex support (for parameters this was already supported).
In addition the nanotec EDS contains octet strings without the leading 0 (as required by 306), I have added support for it as well.
Furthermore I patched the IP mode handling, that prevented the driver to start in my test.

The patches have not been merged upstream yet, since I need to test them first with our devices.

ipa-fmw:
fixed with #75

mistoll:
Great, that was quick.
Unfortunately I have some trouble builiding the indigo-devel branch. So I cannot test it.

The pull request (https://github.com/ipa320/ros_canopen/issues/75) is orthogonal to this issue

floweisshardt:
@mistoll: can you share details on your build errors? e.g. using https://gist.github.com/ and then post a link here?

mistoll:
@ipa-fmw: Sure, here it is: https://gist.github.com/mistoll/2c2e35740f8f19c77d1c
Looks like some boost issue. Boost version is 1.54.0.

mathias-luedtke:
I was not able to reproduce these errors on a clean Ubuntu 14.04 64bit system.
Which system do you use?
Do you have any special compiler settings? (E.g. -std=c++11 does not work).
Have you tried to build it from a clean workspace?

mistoll:
System is Ubuntu 14.04.2 32bit on a 64-bit machine.
Same with clean workspace, no special compiler settings.

mathias-luedtke:
I was able to reproduce these errors in a 32 bit VM, the fix was easy :)

PR #75, opened 2015-03-09
Move ip_mode_sub_mode to configureModeSpecificEntries

mistoll:
The ip_mode_sub_mode entry is specific to interpolated position mode.

The previous implementation breaks compatiblility to devices, which do not support ip mode.

### DIFF
--- a/canopen_402/src/canopen_402/canopen_402.cpp
+++ b/canopen_402/src/canopen_402/canopen_402.cpp
@@ -684,8 +684,6 @@ void Node_402::configureEntries()
   n_->getStorage()->entry(op_mode_display, 0x6061);
   n_->getStorage()->entry(supported_drive_modes, 0x6502);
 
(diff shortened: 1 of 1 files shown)


canopen_402/src/canopen_402/canopen_402.cpp
```diff
@@ -704,6 +702,7 @@ void Node_402::configureModeSpecificEntries()
   }
   if (isModeSupported(Interpolated_Position))
   {
+	n_->getStorage()->entry(ip_mode_sub_mode, 0x60C0);
```
floweisshardt: tabs vs spaces

floweisshardt:
if tabs vs spaces is fixed, then 
visual review: OK

floweisshardt:
@ipa-mdl; pefect!
visual review: OK

mathias-luedtke:
I have already fixed this in my fork (as commented in #74) in the same way, we can merge it now, I will resolve any conflict with my patch later.

PR #76, opened 2015-03-10
fix ambiguous buffer access with 32bit compilers

mathias-luedtke:
Code compiles again on 32 bit system.
(see #74 and https://gist.github.com/mistoll/2c2e35740f8f19c77d1c)


floweisshardt:
visual review: OK

PR #77, opened 2015-03-11
updates from cob4-6

floweisshardt:
- still needs to be tested with other robots
- requires critical review


mathias-luedtke:
cob4-6 version is not suitable for general application -> do not merge.

PR #83, opened 2015-04-07
Improved diagnostics, enabled limit enforcing

mathias-luedtke:
- Warn if not iniatlized
- EMCY error output only if generic errot bit was set
- Enabled limit enforcing again, was commented in previous PR


mathias-luedtke:
tested on cob4-2 and test rig 1: OK

PR #94, opened 2015-05-11
Check if entries are valid before using them

mistoll:
For devices which do not support IP mode, the target_interpolated_position is not valid.


mathias-luedtke:
This code should be obsolete after the 402 refactoring was completed, but it is a reasonable fix and I will merge it as an intermediate solution.

PR #105, opened 2015-07-22
Added validity check to heartbeat_

mistoll:
When using devices, which do not support heartbeat, the old implementation throws InvalidPointer exceptions.

CANOPEN 301, p. 75:
"The implementation of either guarding or heartbeat is mandatory."

Thus those device are standard compliant and should be supported.

### DIFF
--- a/canopen_master/include/canopen_master/canopen.h
+++ b/canopen_master/include/canopen_master/canopen.h
@@ -238,7 +238,7 @@ private:
     PDOMapper pdo_;
 
     boost::chrono::high_resolution_clock::time_point heartbeat_timeout_;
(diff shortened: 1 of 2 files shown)


mistoll:
Ok, that one was just for completeness (Didn't look into getHeartbeatInterval)

mathias-luedtke:
Oh, I meant to rewrite the lines to use getHeartbeatInterval(). This way the implementation details can be altered afterwards.
Could you update them and squash the commits?

mistoll:
Now the other locations also use getHeartbeatInterval().

I also changed the return type of getHeartbeatInterval() from double to uint16_t.

mathias-luedtke:
It is not squashed, but I'll merge it anyway :)
Thanks!

PR #109, opened 2015-07-23
Fix for pluginlib-related crashes on shutdown

### LINKED ISSUE #107: canopen_motor_node does not shutdown properly
The driver might crash due to pluginlib classloader instances that went out of scope.
This might lead to problems with the socketcan kernel driver.

mathias-luedtke:
Since the introduction of the motor layer plugins (dbbc2281458886578250a6514d5ea8bbd1d8b160) the driver crashed silently on shutdown (ctrl+c). However, it was listed in dmesg.

It turns out that the class loader went out scope while motor instances were still be used elsewhere.
This fix introduces a guarded variant of the class loader that keeps the class loader alive until the chain instance was deleted.

closes #107

### DIFF
--- a/canopen_chain_node/include/canopen_chain_node/chain_ros.h
+++ b/canopen_chain_node/include/canopen_chain_node/chain_ros.h
@@ -137,9 +137,36 @@ public:
     virtual ~Logger() {}
 };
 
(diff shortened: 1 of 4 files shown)


mathias-luedtke:
tested on Elmo and Schunk.

Issue #111, opened 2015-08-03
While initializing Target Velocity might be non zero


mistoll:
This assumes PV-Mode but should be valid for other modes, too.

When for the Target Velocity of the motor (on the slave) is non zero, the motor might begin rotating for a short period of time during initialization. This is because the the driver sets the Target Velocity object only after the state Operation Enabled has been entered. This happens independently of the Target Velocity on the master.

The precondition (Target Velocity != 0) might happen during any crash or error condition and cannot be eliminated! 
To reproduce it (hardware dependant!):
1) Start canopen_motor_node (PV-Mode)
2) Init and command the motor to move
3) Stop the driver (Ctrl+C), but do not stop the movement command topic
Now you should have a motor with Target Velocity set to non zero, without a driver running. Continue with:
4) Start canopen_motor_node
5) Call init
Motor should move shortly

IMO this issue is critical, because it's potentially dangerous.

From a more general point of view, this problem could influence every mode. Whenever the motor's object dictionary does not match the dcf default values. But the default values cannot be ensured by a previous driver instance, because a crash might produce an unpredictable motor state. IMO setting each critial object dictionary entry on initialization is not viable solution. I'd suggest to reset each node (NMT reset message) before initializing the node. However that would involve a lag because the node needs to boot.

mathias-luedtke:
We had do remove the deep reset from the driver since it lead to CAN bus problems for Elmo devices.
It can be added again (as an option) after #104 was implemented.

I agree that this issue is critical, I propose to set the Halt bit before the motor is enabled.
This is the default behaviour at start-up, but is currently done shortly (1 sync interval) after the initialization. This delay does not occur on recover.

Can you please test https://github.com/ipa-mdl/ros_canopen/tree/halt-at-init?

In addition driver shutdown should be improved with #109.

Resetting targets to zero is a little bit complicated because of restart cases and the timing would be the same as for the halt bit or even worse.
For now enforcing the target velocity to zero via DCF sounds quite reasonable.
However, this does not work for position modes.

mistoll:
An optional deep reset would be great.

The branch https://github.com/ipa-mdl/ros_canopen/tree/halt-at-init seems to solve the original problem. But while testing it I experiences an even worse problem making it hard to reproduce the problem. Though I managed to run the reproduction steps once, without seeing any motor movement.

[...]

PR #112, opened 2015-08-04
Set Halt bit unless current mode releases it

mathias-luedtke:
Fix for #111


mathias-luedtke:
tested on Elmo and Schunk, the latter might requrire to set drive mode 1 on start-up

Issue #110, opened 2015-08-03
Though not initialized correctly, init service reports success


mistoll:
When a init service call fails a second call to the init service might return successfull while not initializing the driver.

Reproduce like this:
1)
Setup a canopen_motor_node.

2)
Somehow generate a SDO error while calling the service /driver/init().
You might just add "reason = 0x08000000; break;" into sdo.cpp, handleFrame method, Upload*Response cases to force this. (I can provide a patch, if required)
This init call fails (service error, not a error message)

3)
Call /driver/init() a second time. This time the service call returns
{ success: True, message '' }
Despite of displaying "Initializing XXX" the driver does not do much. The controller is not loaded. The motor is still not operational.

mistoll:
In chain_ros.h RosChain::handle_init only canopen::Exception is caught. For other exceptions the state after the exception might be invalid. Maybe there should be a catch all handler?

mathias-luedtke:
Good point, I switched to std::exception to catch them as well.
But I don't think that this a fix for the issue, I guess the exception is caught in between and not handled properly.

mathias-luedtke:
The service call was aborted with an exception because https://github.com/ros-industrial/ros_canopen/commit/00a2c205e6698abaec8825666366be58ecb568ce decoupled TimeoutException from canopen::Exception.

mathias-luedtke:
Should be fixed, can this one be closed?

mistoll:
Not yet. The cleanup might throw an exception and leave the drive in an undefined state.

I added a try/catch around it in #121. It seems to sovle the issue. But it might not reflect the expected behavior!

mathias-luedtke:
Shutdown should never throw. It is a bug if it does, since the clean-up would not finish for all layers.
Which exception do you get? (I bet on the 301 Layer shutdown)
Exception handling must be done at the lowest layer possible, this is currently not implemented for all cases.

This is different to the `init()` function, because a mixed state after a failed init would be cleaned up by `shutdown()` ;)

mistoll:
So what do you think of #123? This should fix the issue and does not add any exception handling.
However there might be other catches.

mathias-luedtke:
I was thinking of catching all exceptions in the Layer base class und pass them as errors, I will prepare a patch for this.
However, I'd really like to know which exception do you get.

mistoll:
The exceptions are (with reproduction hack in place):
First exception (init):

[...]

Issue #64, opened 2015-02-11
Make exception more verbose..


mathias-luedtke:
Exceptions should list the objects they are related to.
In addition exception should only used where necessary.

mathias-luedtke:
Switch to API with return types and catch all exception at lower levels.

mathias-luedtke:
Especially Layer::handleShutdown should never throw an exception.

PR #113, opened 2015-08-04
Verbose exceptions, fix for #110

mathias-luedtke:
As discussed in #110 the exception handling in the service was missing some cases.
In addition all canopen::Exception are verbose now (#64).


mathias-luedtke:
tested on Elmo and Schunk

PR #114, opened 2015-08-05
Stability improvement

mathias-luedtke:
- proper PDO initialization, empty RPDOs throw a TimeoutException
- send SYNC only if needed
- CAN layer recover works


mathias-luedtke:
tested on Elmo and Schunk.

PR #116, opened 2015-08-10
Fix for switching controllers with same mode

mathias-luedtke:
Switching same mode was skipped in creation of switch_map_ and therefore no switch occurred and the drive is stopped eventually.
In addition the creation used the current mode and therefore the map was not generic.

Skipping same-mode-switches is now implemented in the handle layer class.

@ipa-fxm: FYI


mathias-luedtke:
tested on Elmo and Schunk

PR #117, opened 2015-08-11
Reviewed condition variables

mathias-luedtke:
Improved the use of condition variables.
At least the SDO handler was prone to missed condition notifications.


mathias-luedtke:
BufferedReader provides a buffered interface that decouples the socketcan driver from the processing

mathias-luedtke:
Closes #127

mathias-luedtke:
Was tested on Elmo and Schunk successfully.

PR #119, opened 2015-08-12
Improve switch behaviour during emergency stop

### LINKED ISSUE #115: Interface Switching should fail in case of E-Stop
...as it might lead to inconsistent mode states within the chain.

`canSwitch()` should return `false`

mathias-luedtke:
closes #115


mathias-luedtke:
tested on Schunk and Elmo

PR #120, opened 2015-08-13
Enforce target type limits

### LINKED ISSUE #118: Conversion functions should check type bounds
Especially vl_target has only 16 bits and is applied a 64bit double.
The conversion should limit to the type bound and issue a warning.
If wrapping is really needed it should be done with the help of the norm() function.

mathias-luedtke:
Clamping must be done in motor class. 
closes #118


mathias-luedtke:
tested with Schunk.
Logs might flood, but this case should be rare and needs to be handled.

PR #121, opened 2015-08-17
More exception handling

mistoll:
Catching exceptions of shutdown method should solve #110.


mathias-luedtke:
Thanks for your contribution!
If you really need this workaround I will merge it, but unfortunately this is not a proper fix.

mistoll:
No, I understand you do not wan't to catch shutdown exceptions. Let's try something else. 
Sorry, accidentally deleted the branch.

PR #122, opened 2015-08-18
update_ms and sync_ms were mixed up

mistoll:
IMO this makes more sense. And it should do what the documentation says ("The node runs a main loop with CANopen SYNC interval or with an option update interval. " from http://wiki.ros.org/canopen_chain_node?distro=indigo)

### DIFF
--- a/canopen_chain_node/include/canopen_chain_node/chain_ros.h
+++ b/canopen_chain_node/include/canopen_chain_node/chain_ros.h
@@ -429,9 +429,9 @@ protected:
         }
         
         int update_ms = sync_ms;
-        if(sync_ms == 0) nh_priv_.getParam("update_ms", update_ms);
+        nh_priv_.param("update_ms", update_ms, sync_ms);
         if(update_ms == 0){
-            ROS_ERROR_STREAM("Update interval  "<< sync_ms << " is invalid");
+            ROS_ERROR_STREAM("Update interval "<< update_ms << " is invalid");
             return false;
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
I am afraid the documentation is wrong. I have updated it to reflect the current behaviour.
The update interval is coupled to the sync interval, both should match if the SYNC is used, otherwise the timing would be lead to unexpected behaviour.
`update_ms` was introduced to force the loop to wait for the interval.

If both could be set independently:
- If update_ms < sync_ms: the loop might wait for the sync and would exceed update_ms
- If update_ms > sync_ms: some SYNCs might be missed, at least for SimpleMaster.

Perhaps update_ms should be removed in favour of a SyncLayer that does not actually send messages, but waits for the next step.

mistoll:
Maybe you should add this information to the wiki (e.g: Section 3.1.3, ~update_ms). But the current documentation might be enough. 

My original intention was to have a faster update interval than sync, in order to read a value of an async TPDO. Anyhow that's not possible (update_ms < sync_ms ...). So no requirement for update_ms here!

mathias-luedtke:
Hmm, I see.
The driver was designed for synchronous operation, since we need to synchronize multiple axes.
Your use-case should be possible (with your modifications), if you do not want to synchronize your updates to your SYNC messages.

Async PDOs are read immediately, however the update is run on fix intervals coupled to the sync interval.

PR #124, opened 2015-08-19
catch all std exceptions in layer handlers

mathias-luedtke:
Fix for shutdown exceptions (#110)
try catch blocks could be removed from other high-level parts.


mathias-luedtke:
Tested on Elmo.

PR #134, opened 2015-09-18
Reading state_ should be protected by lock

mistoll:


mathias-luedtke:
Thanks, the lock was missing!

PR #142, opened 2015-10-01
added check for CATKIN_ENABLE_TESTING

koellsch:
If you run catkin_make with testing disabled (-DCATKIN_ENABLE_TESTING:=FALSE) you get an error from 'roslaunch-extras.cmake' because it can't find the command 'catkin_run_tests_target'.


floweisshardt:
LGTM

mathias-luedtke:
Good catch, thanks!

PR #143, opened 2015-10-16
Enforce rising edge on fault reset bit on init and recover (Patched)

### LINKED ISSUE #140: Driver might stop working after it failed to recover
First patch (#138) did not work (#139)

mathias-luedtke:
Patched version of #138 
Fixes #140 

@ipa-tif, @ipa-nhg, @ipa-fmw FYI


floweisshardt:
tested on cob4-2: OK

Issue #101, opened 2015-07-06
Make canopen_motor_node accessible via headers


mathias-luedtke:
Move declarations and templates into dedicated headers.

mistoll:
Is this what you expect: https://github.com/mistoll/ros_canopen/commit/d7b4154f7cf4e2c4dbee04b48b227d3cd9905208 ?

I'm going to send a PR as soon as it's ready

mathias-luedtke:
Exactly :D

mathias-luedtke:
solved by #226

PR #144, opened 2015-10-19
Splitted control_node.cpp into control_node.cpp, robot_layer.cpp and …

mistoll:
If you like it, merge it.

Despite being easier to understand there's no improvement here.


mathias-luedtke:
Being easier to understand is a great improvement :smile:
In addition it is a first step to #101 
Thanks a lot!

PR #151, opened 2015-11-11
Support for deprecated PDO remapping if COB-ID is not writable

mathias-luedtke:
Support for old CANopen nodes that do not support deactivating PDOs via COB-ID settings.


mathias-luedtke:
tested with old and new nodes.

PR #152, opened 2015-11-11
Added option to turn off mode monitor

mathias-luedtke:
This is helpful if mode monitoring is not wanted/needed or if object 6061 is not mappable.


mathias-luedtke:
Both options worked with test rig.

PR #150, opened 2015-10-27
Synchronized group switch

mathias-luedtke:
Stops group joints before switch, switches modes and does not forward commands until doSwitch was called.
closes #132 

Should be compatible to https://github.com/ros-controls/ros_control/pull/209

Needs to be tested on multi-joint HW.

PR #153, opened 2015-11-16
deprecated canSwitch

mathias-luedtke:
includes #150 

(will fail until new ros_control version is released)


mathias-luedtke:
rebased, succeeds for shadow-fixed.

PR #156, opened 2016-01-22
Stability fixes

mathias-luedtke:
Minor stability improvements


mathias-luedtke:
Tested with Elmo and Schunk
(needs #150)

Issue #137, opened 2015-09-21
candump to topic  Feature Request


droter:
Hi,

Can ros_canopen publish the candump to a topic?

Thanks,

Matt

mathias-luedtke:
No, this is currently not supported.
Could you give some details on your use case, please?

droter:
I would like to record the CAN traffic to a bag file.  I am trying to compare sensor data with vehicle CAN data.  Is there a different approach that would be better?

For the start the raw candump would work.  It would be nice to publish decoded CAN traffic to a topic.

Thanks,

Matt

mathias-luedtke:
I would just use candump (from can-utils) since this is the only tool that reads the correct timestamps.
In addition you can replay the message afterwards.
Why do you need a rosbag for it?

`canopen_chain_node` can publish simple CANopen objects as topic.

If you need do decode custom CAN messages, then you have to decode them yourself.
A CAN-to-ROS implementation requires 10 lines of code (http://answers.ros.org/question/213645/running-can-radar-on-ros-through-raspberry-pi-2/?answer=215226#post-id-215226).
You can as well filter by ID and decode the messages as needed.

droter:
Thank you for the suggestions.

I was looking at a way to get the candump data publish on a topic so I could record it with the sensor data.  Then all of the data would have the same timestamp and I could play it back.  Can the canopen_chain_node get the candump data?

Matt

droter:
The link to the ros answers is great.  Missed that in searching.  I think CAN for ros is a good thing to work on and appreciate all you guys are doing.

mathias-luedtke:
If you really want to publish CAN raw (can header + payload data) frames then you could inherit from `canopen_chain_node::RosChain` and add your publisher.
But `canopen_chain_node` is not meant to handle CAN frames, it is meant for CANopen networks.

`ros_canopen` does not cover non-CANopen devices, perhaps the CAN package might be moved into a new repo anyway (#103).

iwanders:
Hi,

I'm working with a non-CANopen network and would like to have all CAN messages available a topic as well. This would allow recording of all messages sent over the bus in a bagfile and allow other ROS nodes to convert the messages on the CAN topic into the appropriate types and topics.

[...]

PR #157, opened 2016-01-23
Add socketcan_bridge and can_msgs.

iwanders:
Currently there is no package available in ROS to expose can frames from SocketCAN to a topic or the other way around. This pull request contains the additions which provide this functionality, the development has been discussed in issue #137.

It adds the `can_msgs` package, which holds a message type for a CAN Frame. The `socketcan_bridge` package contains nodes and classes that can bridge frames from SocketCAN to a ROS Topic and vice versa. 

The `socketcan_bridge` builds on the `socketcan_interface` package and does not depend on other `ros_canopen` packages. One change to `socketcan_interface` was required: the `DummyInterface` has been modified in order to facilitate testing the `socketcan_bridge` functionality, test results are unaffected by the changes.

Please review the code, if there are any obstacles to the merge, please let me know. I'll write the wiki pages and send an announcement on the ros-users list when the time comes.

### DIFF
--- a/socketcan_bridge/CMakeLists.txt
+++ b/socketcan_bridge/CMakeLists.txt
@@ -54,3 +54,9 @@ install(DIRECTORY include/${PROJECT_NAME}/
 
 roslint_cpp()
 #for use with catkin_make roslint_socketcan_bridge
+
+if(CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
+  catkin_add_gtest(test_conversion test/test_conversion.cpp)
+  target_link_libraries(test_conversion topic_to_socketcan socketcan_to_topic ${catkin_LIBRARIES})
+endif()
--- a/socketcan_bridge/CMakeLists.txt
+++ b/socketcan_bridge/CMakeLists.txt
@@ -1,17 +1,12 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(socketcan_bridge)
 
-find_package(catkin REQUIRED COMPONENTS
-  socketcan_interface
-  roscpp
-  can_msgs
-  roslint
-)
+find_package(catkin REQUIRED COMPONENTS can_msgs roscpp socketcan_interface)
 
 catkin_package(
   INCLUDE_DIRS include
   LIBRARIES socketcan_to_topic topic_to_socketcan socketcan_interface roscpp can_msgs
-  CATKIN_DEPENDS 
+  CATKIN_DEPENDS can_msgs roscpp socketcan_interface
   DEPENDS Boost
 )
 
@@ -20,11 +15,11 @@ include_directories(include ${Boost_INCLUDE_DIRS} ${catkin_INCLUDE_DIRS})
 
 add_library(socketcan_to_topic src/socketcan_to_topic.cpp)
 target_link_libraries(socketcan_to_topic ${catkin_LIBRARIES} ${Boost_LIBRARIES})
-add_dependencies(socketcan_to_topic can_msgs_generate_messages_cpp)
+add_dependencies(socketcan_to_topic ${catkin_EXPORTED_TARGETS})
 
 add_library(topic_to_socketcan src/topic_to_socketcan.cpp)
 target_link_libraries(topic_to_socketcan ${catkin_LIBRARIES} ${Boost_LIBRARIES})
-add_dependencies(topic_to_socketcan can_msgs_generate_messages_cpp)
+add_dependencies(topic_to_socketcan ${catkin_EXPORTED_TARGETS})
 
 add_executable(socketcan_to_topic_node src/socketcan_to_topic_node.cpp)
 target_link_libraries(socketcan_to_topic_node socketcan_to_topic ${catkin_LIBRARIES} ${Boost_LIBRARIES})
@@ -46,14 +41,15 @@ install(TARGETS
   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
 
 install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
-  FILES_MATCHING PATTERN "*.h")
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
 
-roslint_cpp()
-#for use with catkin_make roslint_socketcan_bridge
 
 if(CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
+  find_package(roslint REQUIRED)
+
+  roslint_cpp()
+  #for use with catkin_make roslint_socketcan_bridge
 
   # unit test for the can_msgs::Frame and can::Frame types.
   catkin_add_gtest(test_conversion test/test_conversion.cpp)
--- a/socketcan_bridge/CMakeLists.txt
+++ b/socketcan_bridge/CMakeLists.txt
@@ -23,11 +23,14 @@ catkin_package(
 include_directories(include ${Boost_INCLUDE_DIRS} ${catkin_INCLUDE_DIRS})
 
 
+
 add_library(socketcan_to_topic src/socketcan_to_topic.cpp)
 target_link_libraries(socketcan_to_topic ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+add_dependencies(socketcan_to_topic can_msgs_generate_messages_cpp)
 
 add_library(topic_to_socketcan src/topic_to_socketcan.cpp)
 target_link_libraries(topic_to_socketcan ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+add_dependencies(topic_to_socketcan can_msgs_generate_messages_cpp)
 
 add_executable(socketcan_to_topic_node src/socketcan_to_topic_node.cpp)
 target_link_libraries(socketcan_to_topic_node socketcan_to_topic ${catkin_LIBRARIES} ${Boost_LIBRARIES})
@@ -57,6 +60,22 @@ roslint_cpp()
 
 if(CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
+
+  # unit test for the can_msgs::Frame and can::Frame types.
   catkin_add_gtest(test_conversion test/test_conversion.cpp)
   target_link_libraries(test_conversion topic_to_socketcan socketcan_to_topic ${catkin_LIBRARIES})
+
+
+  add_rostest_gtest(test_to_socketcan
+                    test/to_socketcan.test
+                    test/to_socketcan_test.cpp)
+  target_link_libraries(test_to_socketcan topic_to_socketcan ${catkin_LIBRARIES})
+
+  add_rostest_gtest(test_to_topic
+                    test/to_topic.test
+                    test/to_topic_test.cpp)
+  target_link_libraries(test_to_topic socketcan_to_topic ${catkin_LIBRARIES})
+
+  
+
 endif()
--- a/socketcan_bridge/CMakeLists.txt
+++ b/socketcan_bridge/CMakeLists.txt
@@ -8,9 +8,6 @@ find_package(catkin REQUIRED COMPONENTS
   roslint
 )
 
-find_package(Boost REQUIRED COMPONENTS system thread)
-
-
 catkin_package(
   INCLUDE_DIRS include
   LIBRARIES socketcan_to_topic topic_to_socketcan socketcan_interface roscpp can_msgs
@@ -18,10 +15,7 @@ catkin_package(
   DEPENDS Boost
 )
 
-
-
 include_directories(include ${Boost_INCLUDE_DIRS} ${catkin_INCLUDE_DIRS})
-
 
 
 add_library(socketcan_to_topic src/socketcan_to_topic.cpp)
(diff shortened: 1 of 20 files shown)


mathias-luedtke:
Thank you very much for your contribution!

I have just three points:
1. `can_msgs` lacks `message_runtime`, please update it with the help of this guide: http://docs.ros.org/indigo/api/catkin/html/howto/format2/building_msgs.html
~~And you do not use `std_msgs`, so please remove the dependency.~~ (Header will be replaced by std_msgs/Header internally)
2. Where did you get the `header.frame_id = "0"` from, I am not aware of this convention.
3. It would be great if `DummyInterface` would actually implement the StateInterface to a certain extent. At least `doesLoopBack()` should reflect the loopback option.

socketcan_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+find_package(catkin REQUIRED COMPONENTS
+  socketcan_interface
+  roscpp
+  can_msgs
+  roslint
+)
+
+catkin_package(
+  INCLUDE_DIRS include
+  LIBRARIES socketcan_to_topic topic_to_socketcan socketcan_interface roscpp can_msgs
```
mathias-luedtke: `socketcan_interface roscpp can_msgs` should go into `CATKIN_DEPENDS`

socketcan_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+  CATKIN_DEPENDS 
+  DEPENDS Boost
+)
+
+include_directories(include ${Boost_INCLUDE_DIRS} ${catkin_INCLUDE_DIRS})
+
+
+add_library(socketcan_to_topic src/socketcan_to_topic.cpp)
+target_link_libraries(socketcan_to_topic ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+add_dependencies(socketcan_to_topic can_msgs_generate_messages_cpp)
```
mathias-luedtke: The recommended way is to depend on `${catkin_EXPORTED_TARGETS}`, see http://docs.ros.org/indigo/api/catkin/html/howto/format2/cpp_msg_dependencies.html

socketcan_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+
+add_library(topic_to_socketcan src/topic_to_socketcan.cpp)
+target_link_libraries(topic_to_socketcan ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+add_dependencies(topic_to_socketcan can_msgs_generate_messages_cpp)
```
mathias-luedtke: same here

socketcan_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+  topic_to_socketcan
+  topic_to_socketcan_node
+  socketcan_bridge_node
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
+
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.h")
```
mathias-luedtke: I had troubles in the past with the `FILES_MATCHING` directive. It is not needed, better remove it.

socketcan_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+
+roslint_cpp()
```
mathias-luedtke: roslint is a test_depend, so maybe the should be moved into the `if(CATKIN_ENABLE_TESTING)` block?
(Including the find package directive)

socketcan_bridge/src/socketcan_to_topic.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  int SocketCANToTopic::setup()
+    {
+      // register handler for frames and state changes.
+      frame_listener_ = driver_->createMsgListener(
+              can::CommInterface::FrameDelegate(this, &SocketCANToTopic::frameCallback));
+
+      state_listener_ = driver_->createStateListener(
+              can::StateInterface::StateDelegate(this, &SocketCANToTopic::stateCallback));
+    };
```
mathias-luedtke, line 50: return is missing.
What should be return? I would stick to bool (false means error) or change it to void.

socketcan_bridge/src/topic_to_socketcan.cpp
```diff
@@ -0,0 +1,91 @@
...
+      can_topic_ = nh->subscribe<can_msgs::Frame>("sent_messages", 10,
+                    boost::bind(&TopicToSocketCAN::msgCallback, this, _1));
+      driver_ = driver;
+    };
+
+  int TopicToSocketCAN::setup()
+    {
+      state_listener_ = driver_->createStateListener(
+              can::StateInterface::StateDelegate(this, &TopicToSocketCAN::stateCallback));
+    };
```
mathias-luedtke, line 46: return is missing.

socketcan_bridge/src/topic_to_socketcan_node.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+  std::string can_device;
+  nh_param.param<std::string>("can_device", can_device, "can0");
+
+  boost::shared_ptr<can::ThreadedSocketCANInterface> driver = boost::make_shared<can::ThreadedSocketCANInterface> ();
+
+  if (!driver->init(can_device, 0))  // initialize device at can_device, 0 for no loopback.
+  {
+    ROS_FATAL("Failed to initialize can_device at %s", can_device.c_str());
+    return -1;
```
mathias-luedtke: please use exit code 1 for failures. For some reason canopen_chain_node and canopen_motor_node use -1, but that is wrong and should be fixed.

gavanderhoorn:
@ipa-mdl wrote:

> 2 . Where did you get the `header.frame_id = "0"` from, I am not aware of this convention.

I'm not sure this is actually still valid / best practice, but it is documented in [std_msgs/Header](http://docs.ros.org/jade/api/std_msgs/html/msg/Header.html):

```
#Frame this data is associated with
# 0: no frame
# 1: global frame
string frame_id
```

mathias-luedtke:
@gavanderhoorn: thanks for the hint, I knew I have seen it somewhere before..

I would suggest to leave the frame empty. I think this is the convention for the joint_states topic.

gavanderhoorn:
Yes, I wouldn't put a `0` or a `1` in there either. Might actually be something to ask on ROS Answers / `ros-users`.

mathias-luedtke:
Related answer by @adolfo-rt: http://answers.ros.org/question/74422/ros-gazebo-plugin-joint-state-topic-frame_id-is-blank/
A CAN bus does not have a single location, so empty should be fine.

iwanders:
Thanks for the extensive review. I'm new to catkin and CMake, the links were much appreciated.

> I had troubles in the past with the FILES_MATCHING directive

That exact line is also produced by calling `catkin_create_pkg`, (although I copied it from `canopen_master`). I read the CMake [install](https://cmake.org/cmake/help/v3.0/command/install.html) documentation; it only serves as a filter. Since there are only .h files in this directory we can do without it indeed.

The header `"0"` was indeed from that the page @gavanderhoorn mentioned. I have changed it to the empty frame as proposed.

I changed the return type of the `setup()` method to void, the `CreateMsgListener` does not provide any indication whether it completed successfully. We could of course check the state of the provided `driver`, but that state is already checked when the `driver::init` when the node is started. To me, it would make no sense to perform these checks twice.

I updated the exit codes, these were indeed copied from the `canopen_chain_node`.

>  It would be great if `DummyInterface` would actually implement the `StateInterface` to a certain extent.

I totally agree, that would also provide a way to test whether the `ROS_INFO` and `ROS_ERROR` messages from the `stateCallback` are working. Is there any standard way for automated testing whether such messages were sent? Or is it acceptable to verify this by hand?

mathias-luedtke:
> That exact line is also produced by calling catkin_create_pkg, (although I copied it from canopen_master).

You are right, the problem was not `FILES_MATCHING` , but the combination with `PATTERN ".svn" EXCLUDE` (as in the catkin_create_pkg template). Sorry for the noise.

> CreateMsgListener does not provide any indication whether it completed successfully.

The current implementation might throw if the system runs out of memory.
You could check that the shared pointer is valid, but this case cannot happen for now.

> Is there any standard way for automated testing whether such messages were sent? Or is it acceptable to verify this by hand?

I don't think that there is a standard way for automated test. You could listen to rosout and check it, but that is not necessary IMHO.

However, you cannot rely on rosout for user to get notified of bus errors, the log output can be muted. Perhaps it would make sense to add diagnostics (http://wiki.ros.org/diagnostic_updater) , but this is really just nice to have.

gavanderhoorn:
@ipa-mdl wrote:

> @iwanders wrote:
> 
> > That exact line is also produced by calling catkin_create_pkg, (although I copied it from canopen_master).
> 
> You are right, the problem was not FILES_MATCHING , but the combination with PATTERN ".svn" EXCLUDE (as in the catkin_create_pkg template). Sorry for the noise.

That filter was really only necessary with old svn versions, as they used to create `.svn` subdirs in all directories of your project. As this is a git project, it is really not needed.

I'd not even bother with it and just remove it.

socketcan_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(socketcan_bridge)
+
+find_package(catkin REQUIRED COMPONENTS can_msgs roscpp socketcan_interface)
+
+catkin_package(
+  INCLUDE_DIRS include
+  LIBRARIES socketcan_to_topic topic_to_socketcan socketcan_interface roscpp can_msgs
```
mathias-luedtke: Please remove `socketcan_interface roscpp can_msgs`, `LIBRARIES` is meant for the library targets of this package.
Other libraries from the dependencies will be added automatically by catkin.

socketcan_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+  CATKIN_DEPENDS can_msgs roscpp socketcan_interface
+  DEPENDS Boost
```
mathias-luedtke: No need to depend on Boost here, since you do not use it directly in your code.
You could keep it, but then you should add the dependency in package.xml as well.

mathias-luedtke:
The tests complete successfully now (they crashed because of the missing return).
And cangen on vcan works as well :)

One last issue:
gtest tests needs this: http://docs.ros.org/indigo/api/catkin/html/howto/format2/gtest_configuration.html

I have found this issue with the help of https://github.com/fkie/catkin_lint

iwanders:
> No need to depend on Boost here, since you do not use it directly in your code.
> You could keep it, but then you should add the dependency in package.xml as well.

It was already in present in the package.xml, but I'm all for removing unnecessary things. I read some more documentation and I believe it is better to leave it out; the additions to the linking were also unnecessary.

> The tests complete successfully now (they crashed because of the missing return).

Excellent, which tests / tool did crash because of the missing return statements? I did not observe any errors from `catkin_make run_tests`, did you use an external tool to perform a static analysis?

> I have found this issue with the help of [https://github.com/fkie/catkin_lint](https://github.com/fkie/catkin_lint)

Thanks, another useful tool; it reports one more issue in the `socketcan_bridge`:

> socketcan_bridge: error: unknown test_depend 'gtest'

But this `test_depend` on `gtest` is actually from the wiki Catkin tutorial page; [line 37 of this example](http://wiki.ros.org/ROS/Tutorials/catkin/CreatingPackage#dependencies_tags). The tests run fine without the `<test_depend>gtest</test_depend>` in the package, I presume you agree with removing it?

mathias-luedtke:
> Excellent, which tests / tool did crash because of the missing return statements? I did not observe any errors from catkin_make run_tests, did you use an external tool to perform a static analysis?

Both add_rostest_gtest targets crashed and rostest just reported "no test output". It took me a while to investigate this. I am using clang++ per default and it seems to handle missing return different from gcc. However, it issued warnings ;)

> But this test_depend on gtest is actually from the wiki Catkin tutorial page; line 37 of this example.

I don't trust the wiki on catkin-related topics anymore, just stick to http://docs.ros.org/indigo/api/catkin/html/howto/format2/gtest_configuration.html and the other useful howtos.

> I presume you agree with removing it?

Yes :)

iwanders:
> I am using clang++ per default and it seems to handle missing return different from gcc. 

Ok, I'll look into using Clang in combination with Catkin.

> I don't trust the wiki on catkin-related topics anymore, just stick to [http://docs.ros.org/indigo/api/catkin/html/](http://docs.ros.org/indigo/api/catkin/html/)

Those pages are indeed a lot more comprehensive, I'll use those as reference from now on.

Thanks for the tips :)

mathias-luedtke:
Everything passes now, great work! Thanks again!
If you have some time left, it would be great if you could add some documentation in the wiki: http://wiki.ros.org/socketcan_bridge.

iwanders:
Great! 

> If you have some time left, it would be great if you could add some documentation in the wiki.

Yes, I'll do that once the spamming issue is resolved on the wiki and account creation is enabled again. I'll also try to get an announcement posted on the users list once the wiki is updated.

PR #164, opened 2016-02-23
hardened code with the help of cppcheck

mathias-luedtke:


fmessmer:
Looks good.
Can you drop a few words about how to use this tool?

mathias-luedtke:
> Can you drop a few words about how to use this tool?

After `catkin_make install` (install space is used to have all headers in one place):

```
cppcheck -I install/include/ -I /opt/ros/indigo/include/ --enable=all --quiet --force src/
```

`src/` would match the full source space and can be limited to subdirectories, e.g. to check per package.
`-I /usr/include/` could be added as well, but I am not sure if this is really helpful.

fmessmer:

PR #166, opened 2016-02-24
stop on internal limit only if limit was not reached before

mathias-luedtke:
Formerly all internal limit errors caused a quickstop, this prevents the recoving.
This patch just sets the error if the drive state is ready for motion AND the internal limit was not reached before.

@ipa-fmw, @ipa-fxm: I was not able to move out of the limit with the velocity interface, but it worked with the trajectory controller on position interface (might jump for 5 degrees).

### DIFF
--- a/canopen_402/include/canopen_402/motor.h
+++ b/canopen_402/include/canopen_402/motor.h
@@ -364,7 +364,7 @@ private:
 
     boost::shared_ptr<Mode> allocMode(uint16_t mode);
 
-    bool readState(LayerStatus &status);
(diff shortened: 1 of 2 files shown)


fmessmer:
LGTM

Successfully moved into joint position limit in velocity mode, stop at limit with diagnostic error state, successfully recovered and able to move away from limit with both `joint_trajectory_controller` as well as `joint_group_velocity_controller`

mathias-luedtke:
rebased and solved merge conflicts

PR #167, opened 2016-02-24
Use fixed control period per default

### LINKED ISSUE #165: Add option for using fixed control periods
The time intervals of the controller execution might vary a lot on non-RT linux.
An option for using the fixed update period should be added.

mathias-luedtke:
On non-real-time systems the calculated control period might fluctuate a lot and is sensible prediction for the next control cycle.

The patch uses the update period (which defaults to the sync period) as control period.
The former behaviour can be restored by setting `use_realtime_period` to true.
However, this is not recommended in general.

Closes #165


fmessmer:
LGTM

PR #168, opened 2016-02-25
added filter chain for state values

mathias-luedtke:
Adds support for filter chains (http://wiki.ros.org/filters) that get applied after unit conversion.

Example config:

``` yaml
nodes:
  rig1_plate_joint:
    ...
    velocity_filters: # or in defaults
    - type: filters/MeanFilterDouble
      name: my_filter
      params:
        number_of_observations: 3
```

### DIFF
--- a/canopen_motor_node/CMakeLists.txt
+++ b/canopen_motor_node/CMakeLists.txt
@@ -7,6 +7,7 @@ find_package(catkin REQUIRED
     canopen_chain_node
     controller_manager
     controller_manager_msgs
+    filters
     hardware_interface
     joint_limits_interface
     joint_limits_interface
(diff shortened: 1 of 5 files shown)


mathias-luedtke:
@ipa-fxm: FYI

fmessmer:
LGTM

Used **without** filter ([driver_config](https://github.com/ipa320/schunk_robots/blob/indigo_dev/schunk_lwa4d/config/arm_driver.yaml)):

no_filter_vel

and **with** filter (config see below):

with_filter_vel

Although I did not test it with the same motion command, the velocity joint_states appear to be a bit smoother (`number_of_observations: 3).

```
name: arm

defaults:
  eds_pkg: schunk_lwa4d
  eds_file: "config/Schunk_0_63.dcf"

  #the following scaling is due to 16bit value range limitation of velocity command in vl mode (2)
  vel_to_device: "rint(rad2deg(vel)*250)"
    "604Csub1": "1" # vl dimension factor numerator
    "604Csub2": "24000" # vl dimension factor denominator
  velocity_filters: # or in defaults
    - type: filters/MeanFilterDouble
      name: mean_filter
      params:
        number_of_observations: 3

nodes:
  arm_1_joint:
    id: 3
  arm_2_joint:
    id: 4
  arm_3_joint:
    id: 5
  arm_4_joint:
    id: 6
  arm_5_joint:
    id: 7
  arm_6_joint:
    id: 8
  arm_7_joint:
    id: 9
```

PR #172, opened 2016-03-09
fix for issue #171

### LINKED ISSUE #171: Fixed control period set to 0.0
After merging https://github.com/ros-industrial/ros_canopen/pull/167, I see the following info message in `rqt_console`:

```
Node: /base/driver
Time: 16:05:34.081296581 (2016-03-04)
Severity: Info
Published Topics: /diagnostics, /rosout

Using fixed control period: 0.000000000

Location:
/u/fxm/git/care-o-bot/src/ros_canopen/canopen_motor_node/src/control_node.cpp:setup:100
```

mathias-luedtke:
As reported #167 had no effect because the duration was read before it got parsed.
closes #171

@ipa-fxm: FYI

### DIFF
--- a/canopen_motor_node/src/control_node.cpp
+++ b/canopen_motor_node/src/control_node.cpp
@@ -95,19 +95,17 @@ public:
 
         ros::Duration dur(0.0) ;
 
-        if(!nh_.param("use_realtime_period", false)){
-            dur.fromSec(boost::chrono::duration<double>(update_duration_).count());
-            ROS_INFO_STREAM("Using fixed control period: " << dur);
-        }else{
-            ROS_INFO("Using real-time control period");
-        }
-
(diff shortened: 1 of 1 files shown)


fmessmer:
@ipa-mdl 
I tested this on the robot and can see now that:
`
[ INFO] [1460465511.287405393]: Using fixed control period: 0.010000000
`
The behavior is still as expected, but I do not know exactly whether I should see a difference and what I can check to verify the fixed-period is actually used.

mathias-luedtke:
Thanks for testing! It still moves, so I will merge this ;)

A difference should be seen if the PC is busy with other processed and cannot meet the loop rate.
The old implementation used the real time difference as input for ControllerManager::update. However, most controller used the period as a prediction for next cycle time. For non-RT system the new default should result in more robust control behaviour. For RT-system it should not make much difference, but can be turned off.

PR #176, opened 2016-05-10
bug in objdict.cpp: add access =="WO" ..

ghost:
when i debug the code, the variable access is always in upper case.
if(access=="wo")....
to
if (access=="wo" || access =="WO")

### DIFF
--- a/canopen_chain_node/src/ros_chain.cpp
+++ b/canopen_chain_node/src/ros_chain.cpp
@@ -473,7 +473,7 @@ bool RosChain::setup(){
 
     std::string hw_id;
     nh_priv_.param("hardware_id", hw_id, std::string("none"));
-
+    hw_id="2";
     diag_updater_.setHardwareID(hw_id);
     diag_updater_.add("chain", this, &RosChain::report_diagnostics);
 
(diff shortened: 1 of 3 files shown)


mathias-luedtke:
The CANopen Electronic data sheet specification  specification states 

```
AccessType: for this object, represented by the following strings ( „ro“ - read only, „wo“ -
write only, „rw“ - read/write, „rwr“ - read/write on process input, „rww“ -
read/write on process output, „const“ - constant value)
```

So uppercase letters do not conform to the standard 306.

For convenience they could be accepted in `ros_canopen`, but not in each branch.
It would be better to convert the string to lowercase and test it afterwards,
This way mixed cases are valid as well.

canopen_chain_node/src/ros_chain.cpp
```diff
@@ -473,7 +473,7 @@ bool RosChain::setup(){
 
     std::string hw_id;
     nh_priv_.param("hardware_id", hw_id, std::string("none"));
-
+    hw_id="2";
```
mathias-luedtke, line 476: this does not belong here

Issue #177, opened 2016-05-19
Won't compile with boost version 1.58


xaedes:
On my system I can't compile canopen_chain_node with boos version 1.58

Boost-Version:

```
dpkg -S /usr/include/boost/version.hpp 
libboost1.58-dev:amd64: /usr/include/boost/version.hpp
```

Compile-Error:

```
In file included from /home/xaedes/gits/fahrrad/workspace/src/canopen_chain_node/src/ros_chain.cpp:3:0:
/home/xaedes/gits/fahrrad/workspace/src/canopen_chain_node/include/canopen_chain_node/ros_chain.h:155:12: error: ‘scoped_ptr’ in namespace ‘boost’ does not name a template type
     boost::scoped_ptr<boost::thread> thread_;
            ^
/home/xaedes/gits/fahrrad/workspace/src/canopen_chain_node/src/ros_chain.cpp: In member function ‘virtual bool canopen::RosChain::handle_init(std_srvs::Trigger::Request&, std_srvs::Trigger::Response&)’:
/home/xaedes/gits/fahrrad/workspace/src/canopen_chain_node/src/ros_chain.cpp:80:5: error: ‘thread_’ was not declared in this scope
     thread_.reset(new boost::thread(&RosChain::run, this));
     ^
/home/xaedes/gits/fahrrad/workspace/src/canopen_chain_node/src/ros_chain.cpp: In member function ‘virtual bool canopen::RosChain::handle_recover(std_srvs::Trigger::Request&, std_srvs::Trigger::Response&)’:
/home/xaedes/gits/fahrrad/workspace/src/canopen_chain_node/src/ros_chain.cpp:118:24: error: ‘thread_’ was not declared in this scope
     if(initialized_ && thread_){
                        ^
/home/xaedes/gits/fahrrad/workspace/src/canopen_chain_node/src/ros_chain.cpp: In member function ‘virtual void canopen::RosChain::handleShutdown(canopen::LayerStatus&)’:
/home/xaedes/gits/fahrrad/workspace/src/canopen_chain_node/src/ros_chain.cpp:156:25: error: ‘thread_’ was not declared in this scope
     if(initialized_ &&  thread_){
                         ^
/home/xaedes/gits/fahrrad/workspace/src/canopen_chain_node/src/ros_chain.cpp: In member function ‘virtual bool canopen::RosChain::handle_shutdown(std_srvs::Trigger::Request&, std_srvs::Trigger::Response&)’:
[...]

PR #178, opened 2016-05-19
added include to <boost/scoped_ptr.hpp>; solving #177

xaedes:
https://github.com/ros-industrial/ros_canopen/issues/177


mathias-luedtke:
Thanks for the patch!

PR #182, opened 2016-06-14
protect MotorChain setup with RosChain lock

### LINKED ISSUE #181: Race condition in MotorChain::setup
There's a race condition in MotorChain::setup.

MotorChain::setup calls RosChain::setup which initalizes the driver services (like init). But the mutex is not blocked by MotorChain::setup. So the service methods are executed after the call to RosChain with the rest of MotorChain::setup not beeing executed yet.

Thus there is a case, where the ControllerManagerLayer is not set up. 
Example log:

> [ INFO] [1465900080.354829180]: Initializing driver
> [ INFO] [1465900085.401722084]: Initializing XXX
> [ INFO] [1465900106.892959379]: Current state: 1 device error: system:0 internal_error: 0 (OK)
> [ INFO] [1465900106.897154048]: Current state: 2 device error: system:0 internal_error: 0 (OK)
> [ INFO] [1465900107.116479495]: Using fixed control period: 0.160000000

Here the last message is from MotorChain::setup, which should be finished before "Initializing driver".

mathias-luedtke:
Should fix #181 

`setup` is replaced by mutex-protected `setup_chain`.

@mistoll: please review

### DIFF
--- a/canopen_chain_node/include/canopen_chain_node/ros_chain.h
+++ b/canopen_chain_node/include/canopen_chain_node/ros_chain.h
@@ -194,9 +194,10 @@ protected:
     bool setup_nodes();
     virtual bool nodeAdded(XmlRpc::XmlRpcValue &params, const boost::shared_ptr<canopen::Node> &node, const boost::shared_ptr<Logger> &logger);
     void report_diagnostics(diagnostic_updater::DiagnosticStatusWrapper &stat);
+    virtual bool setup_chain();
 public:
     RosChain(const ros::NodeHandle &nh, const ros::NodeHandle &nh_priv);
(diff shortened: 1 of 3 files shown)


mistoll:
Simple and fixes #181 

Log of the correct behavior (with 9e626bfc5bb69474468467ae4f8a846e653e1a68 in place):

> [ INFO] [1465921618.422371345]: Initializing XXX => Service method called
> [ INFO] [1465921619.123927591]: Using fixed control period: 0.160000000 => Inside setup
> [ INFO] [1465921619.129103373]: Current state: 1 device error: system:0 internal_error: 0 (OK) => Service method enters mutex

Issue #183, opened 2016-06-29
One object can't be mapped to multiple PDO


JuliusGel:
At the moment one object can only be mapped in a single PDO. If an object is configured to be sent by multiple PDOs, ObjectStorage::map method will make it so that this object is only updated with the last PDO object. Ideally it should be possible to update the object from multiple PDOs. Alternatively the user should at least be notified that one object can only be mapped in single PDO.

mathias-luedtke:
I am afraid the current design cannot support multiple mappings.
There is only a single callback that is used for synchronization.

PR #185, opened 2016-06-30
check for doubled PDO mappings

mathias-luedtke:
As pointed out in #183 PDOs cannot be mapped twice. This patch checks for this case and stops further processing.

@JuliusGel: Can you test this fix, please.
I don't have access to a CANopen set-up right now, and it is triggered only in 301 init.

### DIFF
--- a/canopen_master/include/canopen_master/canopen.h
+++ b/canopen_master/include/canopen_master/canopen.h
@@ -12,6 +12,7 @@
 #include <boost/thread/condition_variable.hpp>
 #include <boost/chrono/system_clocks.hpp>
 #include <boost/lexical_cast.hpp>
+#include <boost/function.hpp>
 
 namespace canopen{
 
@@ -87,9 +88,14 @@ class PDOMapper{
     class PDO {
     protected:
         void parse_and_set_mapping(const ObjectStorageSharedPtr &storage, const uint16_t &com_index, const uint16_t &map_index, const bool &read, const bool &write);
+        boost::function<void()> unmap;
         can::Frame frame;
         uint8_t transmission_type;
         std::vector<BufferSharedPtr>buffers;
+    public:
+        ~PDO(){
+            if(unmap) unmap();
+        }
     };
 
     struct TPDO: public PDO{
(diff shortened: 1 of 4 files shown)


JuliusGel:
We have tested this PR. This correctly prevents one object mapping in multiple PDO.

JuliusGel:
There is an issue with this fix. If you call /driver/init then call /driver/shutdown and call /driver/init again there will be an exception thrown about PDO mapped twice.

mathias-luedtke:
mapping needs to be reset in
https://github.com/ros-industrial/ros_canopen/blob/indigo-devel/canopen_master/src/pdo.cpp#L173
and
https://github.com/ros-industrial/ros_canopen/blob/indigo-devel/canopen_master/src/pdo.cpp#L186

mathias-luedtke:
> There is an issue with this fix. 

Should be rescolved by auto-unmap.

mathias-luedtke:
@JuliusGel: does the new version work for you?

JuliusGel:
@ipa-mdl : Hi, unfortunately we had made a decision not use ros_canopen due to hardware issues. I therefore am not able to test this anymore. I could only provide a dcf that was failing if that is of any interest.

mathias-luedtke:
> I could only provide a dcf that was failing if that is of any interest. 

This would be great :)

JuliusGel:
I am attaching the eds file that had this problem.
We were using it with Trinamic motor PD-1160. Hope this is useful.
[TMCM-1160_V318.zip](https://github.com/ros-industrial/ros_canopen/files/560443/TMCM-1160_V318.zip)

jacknlliu:
@ipa-mdl What's the status of this PR? We have the same issue, is there a solution for this issue? Can ros_canopen support multiple-pdos mapping now?

mathias-luedtke:
@jacknlliu: It must be rebased and tested.

jacknlliu:
@ipa-mdl actually this PR only check the mapping and don't make ros_canopen work with   this situation. If you have solutions to make doubled PDOs work with ros_canopen, we can test it.

mathias-luedtke:
The current design does not allow ambiguous PDO mappings.
This PR will just make it report the problem.

Issue #161, opened 2016-02-22
Jade/Kinetic adaptions


mathias-luedtke:
Mode switching needs to support mixed joint settings.

mgruhler:
FYI, there has been a C++ API breaking change in `ros_control` for kinetic: https://github.com/ros-controls/ros_control/commit/afaf9403d1daf6e7f0a93e4a06aa9695e2883632

mathias-luedtke:
That's what I meant with  "mixed joint settings" ;)

mgruhler:
Ooops :-D

mistoll:
I'd like to provide a pr for this.
It could be solved with a simple additional for loop over claimed_resources. Are there any doubts about that?
How do you want to select between the old and the new api? (Separate branch, conditional compilation)

mathias-luedtke:
> How do you want to select between the old and the new api?

separate branch, `jade-devel` for jade and kinetic.

> It could be solved with a simple additional for loop over claimed_resources. Are there any doubts about that?

This works only for controllers without mixed modes.
However, it should be okay to just loop over the claimed_resources, but fail if more than one resource type is needed. (-> this supports all indigo-based controllers)

Mixed mode settings should be supported as well, but can be added later (#197).

PR #196, opened 2016-11-18
ros_control adaption

mistoll:
Fix for #161 

This should NOT be merged to the indigo-devel branch, because it's not compatible with the indigo version of ros_control.

Not tested.
Open for discussion.


mathias-luedtke:
Thanks!
This should do the trick for single-mode cases.

I will not merge this PR, but prepare a new one with an additional single-mode check on top of your commit.

PR #199, opened 2016-11-18
Jade fixes and single-mode check

mathias-luedtke:
contains #198


mistoll:
Tested it on our kinetic system. No Problems, so far.

mathias-luedtke:
If it works with kinetic, jade should be fine as well.
Not sure if we need a specific kinetic branch.

mistoll:
I dont't have a jade system, so can't say anything about that.

IMO you should keep it simple and only create a kinetic branch, if you have to.

floweisshardt:
kinectic is the official LTS of ROS, so it makes sense for me to have a kinectic branch (and probably skip support for the jade branch)

mathias-luedtke:
I will keep the `jade-devel` as long as it works for jade and kinetic and will branch as soon as a breaking change occurs.

At some point `jade-devel` will become the master branch.
TODO: add hint on compatibility to readme

Issue #71, opened 2015-02-20
Refactor Sync mechanism


mathias-luedtke:
The current implementation might not work properly with multiple chains.

Issues:
- shared memory clean-up
- permission settings between multiple users
- crash recovery of locks (#53)

To be refactored:
- Book keeping of processes
- Book keeping of node IDs
- Master mutex should not be locked all the time
- Crashed masters should be detected
- Recovery
- Diagnostics for missed intervals etc.
- plugin interface for easy exchange of sync implementation

mathias-luedtke:
Handle cases with update rate != sync rate (see #122)

mathias-luedtke:
This is mostly solved by #201

PR #201, opened 2016-12-10
Added support for external sync providers

mathias-luedtke:
addresses #71 

This patch contians a new sync plugin (`canopen::ExternalMaster::Allocator`) for external sync sources and a BCM-based sync provider.
The sync provider monitors NMT and Heartbeat messages to decide when to start and stop the sync.

The CAN BCM generates the SYNC in kernel space, so it should have less jitter.

Usage:
```
rosrun canopen_master canopen_bcm_sync DEVICE PERIOD_MS HEADER [OVERFLOW] [+ID*] [-ID*] [--]
```
The `+-ID` option can be used for white/black listing.

As an alternative `canopen_sync_node` with ROS parameter suppport is available (`canopen_chain_node` syntax).

### DIFF
--- a/canopen_chain_node/CMakeLists.txt
+++ b/canopen_chain_node/CMakeLists.txt
@@ -62,10 +62,20 @@ target_link_libraries(canopen_chain_node
   ${catkin_LIBRARIES}
 )
 
+# canopen_sync_node
+add_executable(canopen_sync_node
+  src/sync_node.cpp
+)
+target_link_libraries(canopen_sync_node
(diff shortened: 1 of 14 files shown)


mathias-luedtke:
The BCM sync introduces even less jitter than the (default) SimpleMaster.

mathias-luedtke:
tested with Schunk and Elmo hardware.
3c3de07e1f56d8870fc7bcf1a3562d43fc2b4ba7 might influence IPC/SHM based sync masters, they got deprecated

PR #202, opened 2016-12-11
Object logging in diagnostics

mathias-luedtke:
Refactored existing code for object logging to make it work again.

Each node can take a `log` (all cases) ,`log_warn` (warn and error), or `log_error` (only error) list which adds diagnostics entries.
Syntax is the same as for `publish`, with ! forcing an update.


mathias-luedtke:
push some fixes and tested on Schunk.

PR #203, opened 2016-12-13
backport: check if hardware interface matches mode

mathias-luedtke:
backport of d8f047fab84268efd2ecfbeab52961379ce2e34b from jade to indigo


mathias-luedtke:
tested with schunk

PR #208, opened 2016-12-21
removed IPC/SHM based sync masters

mathias-luedtke:
This PR removes the deprecated IPC/SHM based sync masters.
Should be merged after the next release.


mathias-luedtke:
I will merge this after the final `jade` release.

PR #220, opened 2017-02-06
Improve recovery behaviour

### LINKED ISSUE #217: Failed init in chain node result in infinite loop
#201/#207 introduces a bug that might lead to an infinite loop if init fails.
The run thread does not get stopped correctly and just gets detached. This results in a high load if the external sync provider is used.

mathias-luedtke:
Fixes #217 for real..
Prevents race conditions if errors occur while init and recover.


mathias-luedtke:
tested with cob4-2

PR #224, opened 2017-03-07
do not call handleReadread in HandleLayer::handleRecover

mathias-luedtke:
this prevents a race condition, it is not needed anyway.


floweisshardt (approved):
tested on cob4 base did not carsh in >20h of testing.

PR #226, opened 2017-03-09
Make canopen_motor_node reusable

### LINKED ISSUE #101: Make canopen_motor_node accessible via headers

mistoll:
There are two changes here:

https://github.com/mistoll/ros_canopen/commit/1de23dc1807473419dbe5c7bef12e1e1f15cf0a4
This commit does only split the headers and cpp files into multiple units. 
It's mostly cut & paste. However the classes were moved to the canopen namespace.
Also a library and exports were added to CMakeLists.txt
Closes #101 

https://github.com/mistoll/ros_canopen/commit/2c774e36183cd75defd7b310900e978172056c7f
The previous commit allows using the code as a library in other packages. However the tight coupling between HandleLayer and RobotLayer reduces the cases where it is resonable to do so. Thus a pure virtual class HandleLayerBase was introduced. That allows the implementation of custom HandleLayers, which do not use the MotorBase class.
This is just a proposal. I chose this solution because it's very simple and has low impact. However there might be way better ideas, especially with the view on the whole project.


mathias-luedtke:
Thanks a lot!
This will take some time to review.

> Thus a pure virtual class HandleLayerBase was introduced.

Sound reasonable :)

canopen_motor_node/CMakeLists.txt
```diff
@@ -22,41 +22,76 @@ find_package(Boost REQUIRED
...
 # canopen_motor_node
 add_executable(canopen_motor_node
-  src/control_node.cpp
-  src/robot_layer.cpp
+  src/canopen_motor_chain_node.cpp
 )
 target_link_libraries(canopen_motor_node
+  canopen_motor
   ${catkin_LIBRARIES}
-  ${MUPARSER_LIBRARIES}
```
mathias-luedtke, line 40: why did you remove MUPARSER here, but kept catkin?

canopen_motor_node/src/interface_mapping.h
```diff
@@ -0,0 +1,48 @@
+
```
mathias-luedtke, line 1: I am not sure if this should be part of the public API

mathias-luedtke:
Just some minor questions :)

Great job :+1: 

Next step: I will test this with ~~kinetic~~jade

canopen_motor_node/CMakeLists.txt
mistoll: I've splitted canopen_motor_node into canopen_motor library and canopen_motor_node executable. MUPARSER is linked into canopen_motor library. 
IMO the executable canopen_motor_node does not require MUPARSER.

canopen_motor_node/src/interface_mapping.h
mistoll: It's not. The header is placed into the src folder, not the include folder.
I've used that method for private headers in some packages and it works great. However I haven't seen that in the ros_canopen packages, yet. It might not fit into the coding guidelines.

mathias-luedtke: Ah, that's fine! sorry for the noise!

> I've used that method for private headers in some packages and it works great

Me too :)

canopen_motor_node/CMakeLists.txt
mathias-luedtke: I just realised that canopen_motor_node might need additional lib froms catkin, that have been stripped from canopen_motor.

mathias-luedtke (approved):
LGTM

I will run a prerelease test, but I don't expect any problems.

Will get merged after a hardware test.

mathias-luedtke:
I have run a small test, everything worked as expected :+1: 
I will trigger a release.

PR #234, opened 2017-04-27
Feature: Reset errors before recover

mathias-luedtke:
The feature introduces a flag to clear errors before recovering.
It might be needed for old Elmo controllersm but is turned off by default.

To turn it off `reset_errors_before_recover` should be set to "true" in private namespace.

@ipa-jba: Please test with new option
@ipa-nhg: Please test that recover works without setting this option on recent hardware


jabbenseth:
only to document our in person conversation results: 
- this PR solves the problems we had on raw3-3

fmessmer:
also works on `cob4-9`...

PR #239, opened 2017-06-07
error instead of warning for motor not enabled

floweisshardt:
fixes issue with auro_recover which only recovers on error not on failure.

@ipa-mdl: please review. Is setting to warning only done by purpose? what's the rationale behind that?
If you approve, I'll open a PR to jade too.


mathias-luedtke:
> fixes issue with auro_recover which only recovers on error not on failure.

What do you mean by `failure`?

>  Is setting to warning only done by purpose? what's the rationale behind that?

It is a warning only as these states are part of the regular function (e.g. while init and recover). Error should only be sent in case of emergency errors, i.e. if chain needs to be recovered.

mathias-luedtke:
I don't want to merge this. Some drives (including the [cob4 components in twist mode](https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_hardware_config/robots/cob4-2/config/torso_driver.yaml#L13)), require the state machine to switch to other states, most likely  `Switched_On`. This PR would lead to spurious errors.

fmessmer:
why close? do we have an alternative yet?

in the first place, what was the problem again that occured without this fix...just trying to understand as we currently use this feature branch in our cob scenarios

mathias-luedtke:
> why close? do we have an alternative yet?

"This PR would lead to spurious errors." (see above)
It is not a proper fix for the (not yet reported) failure.

PR #243, opened 2017-07-26
make can::Header/Frame::isValid() const

### LINKED ISSUE #242: [socketcan_interface] interface.h isValid not const
We can not call the isValid function on a const can::Frame.
What about adding a const to this function which only read the can::Frame data ?

mathias-luedtke:
fixes #242

### DIFF
--- a/socketcan_interface/include/socketcan_interface/interface.h
+++ b/socketcan_interface/include/socketcan_interface/interface.h
@@ -21,7 +21,7 @@ struct Header{
     unsigned int is_rtr:1; ///< frame is a remote transfer request
     unsigned int is_extended:1; ///< frame uses 29 bit CAN identifier 
     /** check if frame header is valid*/
-    bool isValid(){
+    bool isValid() const{
         return id < (is_extended?(1<<29):(1<<11));
     }
     /** constructor with default parameters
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
~~I guess this is an ABI break..
Perhaps it makes sense to just keep the non-const function as well.~~
No: https://travis-ci.org/ipa-mdl/ros_canopen/jobs/257941917

gavanderhoorn:
<del>Off-topic, but: that ABI test looks interesting. Can you tell me how you implemented / configured that? I see no mention of it in the `.travis.yml` of this repository.</del>

<del>O wait, I'm guessing that you added it in 7b707f603b68ea74d2b7ec29edf36e6f16a219e7. Is that a recent addition to `industrial_ci`? That seems like an awesome piece of functionality.</del>

and I just found ros-industrial/industrial_ci#199.

mathias-luedtke:
Actually, I am not really sure if the ABI checks hold for this specific PR as the change affect the header only. I will prepare a test locally to check this.

Update: seems to be okay.

mathias-luedtke:
@vincentrou: Do you need it for `indigo`? Since I am not 100% sure about the API/ABI break, I would prefer to keep it as-is.

I will release hust before the final `jade` sync.

Issue #230, opened 2017-04-04
Add CAN ID and mask to socketcan_bridge


9a3eedi:
It would be great if socketcan_bridge had the ability to only publish messages that has a certain CAN ID, specified by a rosparam. I can't seem to find any way of doing it other than creating a new node that filters out messages coming in/out from the bridge.

mathias-luedtke:
If you just want to filter a single CAN ID, you can add it [here](https://github.com/ros-industrial/ros_canopen/blob/indigo-devel/socketcan_bridge/src/socketcan_to_topic.cpp#L46) easily .Otherwise more code needs to be written.

Yor're welcome to provide a pull request :)

PR #245, opened 2017-09-11
can id pass filter for socketcan bridge

benmaidel:
Added feature to define a list of can ids which should be forwarded from can bus to ros. All other can ids are ignored. If the list is empty all messages get forwarded to ros.
```
Param: can_ids
Type: list<string>
Examples: 
- ['0x123', '0x321', '0x444']
- ['123', '321, '444']
```
solves https://github.com/ros-industrial/ros_canopen/issues/230


mathias-luedtke:
After our discussion today, I came up with https://github.com/ipa-mdl/ros_canopen/tree/feature/can-filter as a general solution with masks and ranges

mathias-luedtke:
closed in favor of #250

PR #263, opened 2018-01-17
find and link the thread library properly

bulwahn:


mathias-luedtke:
Thanks (again) fot the fix!
I wonder if it makes sense to use the [code from roscpp](https://github.com/ros/ros_comm/blob/lunar-devel/clients/roscpp/CMakeLists.txt#L41) to export the setting to other packages.

bulwahn:
I am sorry, but I don't know this good enough to really judge if the more complex solution from roscpp would be a better solution. I only came across this one issue in meta-ros and that was easily resolved with this minimal solution. Possibly, it is best to wait for another user to report if there is a further problem down the road rather than just proactively integrating some more code where we don't know yet if that issue even can happen.

mathias-luedtke (approved):
I think it might be better to link against the [`Threads::Threads` target](https://cmake.org/cmake/help/v3.1/module/FindThreads.html).
Unfortunately, Debian jessie ships only with CMake 3.0.2.

Your patch seems to be the most-portable way to fix this issue.

I will merge it after you have removed the extra commit.

bulwahn:
I just rebased it on top of the current kinetic-devel. It should be now ready to be merged.

PR #266, opened 2018-01-31
Removing CAN_ERR_LOSTARB from the list of errors. Reasoning is, that …

DasRoteSkelett:
…for example sja1000 based cards issue this error frame, even though, the message is send over the can bus, as lost Arbitrations are a rather common thing to happen on a CAN bus. Since this is rather a statistical hint, not an error per se, it is the wrong behavior to put the socketcan_interface into error state, which can only be resolved by "recover()", i.e. close and open the driver again.

I can confirm this behavior for an sja1000 based card. There are (only) 3 more drivers, that use this error frame at all (rcar, sun4i, xilinx). Have not tested those.


DasRoteSkelett:
I see that the same topic came up a couple of days ago in another PR

PR #277, opened 2018-05-04
Added state_switch_timeout parameter to motor.

agutenkunst:
Enables users to set the timeout e.g.
<pre>
defaults: # optional, all defaults can be overwritten per node
  eds_pkg: my_config_package # optional package  name for relative path
  eds_file: "my_config.dcf" # path to EDS/DCF file
  <b>motor_layer:
    state_switch_timeout: 15</b>
    "1016sub1" : "0x7F0064" # heartbeat timeout of 100 ms for master at 127
    "1017": "100" # heartbeat producer
</pre>
currently defaults to 5s to ensure consistent behaviour on deployed systems.


mathias-luedtke (approved):
Looks good, thanks!

PR #286, opened 2018-06-01
Improve logging system

mathias-luedtke:
This PR implements logging based on `console_bridge`.

* deprecated `LOG` macro
* added ROS bindings to all nodes

### DIFF
--- a/canopen_402/include/canopen_402/motor.h
+++ b/canopen_402/include/canopen_402/motor.h
@@ -134,7 +134,7 @@ public:
     virtual bool start() = 0;
     virtual bool read(const uint16_t &sw) = 0;
     virtual bool write(OpModeAccesser& cw) = 0;
-    virtual bool setTarget(const double &val) { LOG("not implemented"); return false; }
+    virtual bool setTarget(const double &val) { ROSCANOPEN_ERROR("not implemented"); return false; }
     virtual ~Mode() {}
 };
 typedef std::shared_ptr<Mode> ModeSharedPtr;
@@ -149,7 +149,7 @@ public:
     T getTarget() { return target_; }
     virtual bool setTarget(const double &val) {
         if(std::isnan(val)){
-            LOG("target command is not a number");
+            ROSCANOPEN_ERROR("target command is not a number");
             return false;
         }
 
@@ -162,15 +162,15 @@ public:
             target_= numeric_cast<T>(val);
         }
         catch(negative_overflow&) {
-            LOG("Command " << val << " does not fit into target, clamping to min limit");
+            ROSCANOPEN_WARN("Command " << val << " does not fit into target, clamping to min limit");
             target_= std::numeric_limits<T>::min();
         }
         catch(positive_overflow&) {
-            LOG("Command " << val << " does not fit into target, clamping to max limit");
+            ROSCANOPEN_WARN("Command " << val << " does not fit into target, clamping to max limit");
             target_= std::numeric_limits<T>::max();
         }
         catch(...){
-            LOG("Was not able to cast command " << val);
+            ROSCANOPEN_ERROR("Was not able to cast command " << val);
             return false;
         }
 
--- a/canopen_402/include/canopen_402/motor.h
+++ b/canopen_402/include/canopen_402/motor.h
@@ -134,7 +134,7 @@ public:
     virtual bool start() = 0;
     virtual bool read(const uint16_t &sw) = 0;
     virtual bool write(OpModeAccesser& cw) = 0;
-    virtual bool setTarget(const double &val) { ROSCANOPEN_ERROR("not implemented"); return false; }
+    virtual bool setTarget(const double &val) { ROSCANOPEN_ERROR("canopen_402", "Mode::setTarget not implemented"); return false; }
     virtual ~Mode() {}
 };
 typedef std::shared_ptr<Mode> ModeSharedPtr;
@@ -149,7 +149,7 @@ public:
     T getTarget() { return target_; }
     virtual bool setTarget(const double &val) {
         if(std::isnan(val)){
-            ROSCANOPEN_ERROR("target command is not a number");
+            ROSCANOPEN_ERROR("canopen_402", "target command is not a number");
             return false;
         }
 
@@ -162,15 +162,15 @@ public:
             target_= numeric_cast<T>(val);
         }
         catch(negative_overflow&) {
-            ROSCANOPEN_WARN("Command " << val << " does not fit into target, clamping to min limit");
+            ROSCANOPEN_WARN("canopen_402", "Command " << val << " does not fit into target, clamping to min limit");
             target_= std::numeric_limits<T>::min();
         }
         catch(positive_overflow&) {
-            ROSCANOPEN_WARN("Command " << val << " does not fit into target, clamping to max limit");
+            ROSCANOPEN_WARN("canopen_402", "Command " << val << " does not fit into target, clamping to max limit");
             target_= std::numeric_limits<T>::max();
         }
         catch(...){
-            ROSCANOPEN_ERROR("Was not able to cast command " << val);
+            ROSCANOPEN_ERROR("canopen_402", "Was not able to cast command " << val);
             return false;
         }
 
--- a/canopen_chain_node/src/ros_chain.cpp
+++ b/canopen_chain_node/src/ros_chain.cpp
@@ -593,7 +593,7 @@ RosChain::~RosChain(){
         LayerStatus s;
         halt(s);
         shutdown(s);
-    }catch(...){ LOG("CATCH"); }
+    }catch(...){ ROS_ERROR("CATCH"); }
 }
 
 }
--- a/canopen_master/src/emcy.cpp
+++ b/canopen_master/src/emcy.cpp
@@ -42,7 +42,7 @@ struct EMCYmsg{
 
 void EMCYHandler::handleEMCY(const can::Frame & msg){
     EMCYmsg::Frame em(msg);
-    ROSCANOPEN_ERROR("EMCY: " << msg);
+    ROSCANOPEN_ERROR("canopen_master", "EMCY received: " << msg);
     has_error_ = (em.data.error_register & ~32) != 0;
 }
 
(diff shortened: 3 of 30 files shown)


mathias-luedtke:
@jschleicher: did you ever test this? I think it's about time to merge this :)

jschleicher:
I'll give it a try later today.

canopen_master/src/emcy.cpp
```diff
@@ -42,7 +42,7 @@ struct EMCYmsg{
 
 void EMCYHandler::handleEMCY(const can::Frame & msg){
     EMCYmsg::Frame em(msg);
-    LOG("EMCY: " << can::tostring(msg, false));
+    ROSCANOPEN_ERROR("EMCY: " << msg);
```
jschleicher: Please use tostring to output hex codes (much more usable):
`[ERROR] [1562077572.093793150]: 83#0000000000000000`
```suggestion
    ROSCANOPEN_ERROR("EMCY: " << can::tostring(msg, false).c_str());
```

canopen_chain_node/src/ros_chain.cpp
```diff
@@ -593,7 +593,7 @@ RosChain::~RosChain(){
         LayerStatus s;
         halt(s);
         shutdown(s);
-    }catch(...){ LOG("CATCH"); }
+    }catch(...){ ROS_ERROR("CATCH"); }
```
jschleicher, line 596: What is the logic here? canopen_chain_node, canopen_motor_node and socketcan_bridge use `ROS_*`-macros directly while the ros-agnostic packages use rosconsole's `ROSCANOPEN_*` macros? Please add a sentence to CONTRIBUTING.md or similar explaining the idea to other developers!

jschleicher (changes requested):
Using the ros logging system gives convenient timestamps etc. for every message, so a general :+1: for this. 
See inline comments.

canopen_master/src/emcy.cpp
mathias-luedtke: Can you post an example output without your fix?

canopen_chain_node/src/ros_chain.cpp
mathias-luedtke: >  canopen_chain_node, canopen_motor_node and socketcan_bridge use ROS_*-macros directly while the ros-agnostic packages use rosconsole's ROSCANOPEN_* macros

Exactly.

> Please add a sentence to CONTRIBUTING.md or similar explaining the idea to other developers!

Good idea.

canopen_master/src/emcy.cpp
jschleicher: Example output without my fix (your version) is a plain decimal:
from rosout.log ```1562076301.787823008 ERROR /prbt/driver [.../emcy.cpp:45()] [topics: /rosout, /diagnostics, ...] 131```

jschleicher: By the way: why is the "EMCY: " string lost in both cases?

mathias-luedtke: Both issues are bugs, I am on it.

mathias-luedtke: I think I fixed it ^^

socketcan_interface/include/socketcan_interface/socketcan.h
```diff
@@ -200,7 +203,7 @@ class SocketCANInterface : public AsioDriver<boost::asio::posix::stream_descript
                 input_.id = frame_.can_id & CAN_EFF_MASK;
                 input_.is_error = 1;
 
-                LOG("error: " << input_.id);
+                ROSCANOPEN_ERROR("error: " << input_.id);
```
jschleicher: Outputs only the id:
```
[ERROR] [1562162011.086172652]: 4
```

canopen_master/src/emcy.cpp
jschleicher: Tried 7e45ccdce754eff61db998436cc39b7bce06c5ba and the output now reads:
```
[ERROR] [1562161991.081027610]: 83#0000000000000000
```
Note: "EMCY: " is still missing.

jschleicher (commented):
The strings are swallowed somewhere...

mathias-luedtke:
if I remove `REGISTER_ROSCONSOLE_BRIDGE`, everything gets printed to stdout.

mathias-luedtke:
It's not a bug, it's a ["feature"](https://github.com/ros/rosconsole_bridge/blob/kinetic-devel/src/bridge.cpp#L52) :-/

mathias-luedtke:
@jschleicher: the work-around should the filtering
I don't want to merge it this way. Instead I will prefix all message, but this takes some time to implement.

canopen_402/include/canopen_402/motor.h
```diff
@@ -134,7 +134,7 @@ class Mode {
     virtual bool start() = 0;
     virtual bool read(const uint16_t &sw) = 0;
     virtual bool write(OpModeAccesser& cw) = 0;
-    virtual bool setTarget(const double &val) { LOG("not implemented"); return false; }
+    virtual bool setTarget(const double &val) { ROSCANOPEN_ERROR("canopen_402", "Mode::setTarget implemented"); return false; }
```
jschleicher: Left out "not" intentionally?
```suggestion
    virtual bool setTarget(const double &val) { ROSCANOPEN_ERROR("canopen_402", "Mode::setTarget not implemented"); return false; }
```

socketcan_interface/include/socketcan_interface/socketcan.h
```diff
@@ -200,7 +203,7 @@ class SocketCANInterface : public AsioDriver<boost::asio::posix::stream_descript
                 input_.id = frame_.can_id & CAN_EFF_MASK;
                 input_.is_error = 1;
 
-                LOG("error: " << input_.id);
+                ROSCANOPEN_ERROR("socketcan_interface", "error frame: " << input_.id);
```
jschleicher: Why do you limit the output to the id? The subsequent message in can_layer prints the whole frame:
```
[ERROR] [1562238561.394633541]: error frame: 4
[ERROR] [1562238561.394850816]: Received error frame: 20000000#0030000000000000
```

jschleicher (commented):
Tested 8ab0afa80713d8f1d327da83ecc5d6e84c45c6ff; output of the strings works now. See more inline comments.

canopen_402/include/canopen_402/motor.h
mathias-luedtke: Ehm, no ;)

socketcan_interface/include/socketcan_interface/socketcan.h
mathias-luedtke: I just batch-replaced, but it sounds resonable to print the full frame :)

mathias-luedtke: What confuses my right now: shouldn't the second line be `20000004`?

mathias-luedtke: IMHO the first message could get dropped completely

jschleicher: actually I disconnected the CAN-bus, so the error is "could not reset node '3'".
Candump shows nothing besides 
```(1562591423.558640)  can0  TX - -  000   [2]  82 03```
and thus I am unsure, what the error telegram with ID 4 says. Is there an internal error message generated by socketcan?

Comment (jschleicher) [reply socketcan_interface/include/socketcan_interface/socketcan.h]: Yes I agree, that the first error message could be dropped. It is just duplicated information (once the second output is interpreted correctly).

File: socketcan_interface/include/socketcan_interface/socketcan.h [hunk 6]
```diff
@@ -200,7 +203,7 @@ class SocketCANInterface : public AsioDriver<boost::asio::posix::stream_descript
                 input_.id = frame_.can_id & CAN_EFF_MASK;
                 input_.is_error = 1;
 
-                LOG("error: " << input_.id);
+                ROSCANOPEN_ERROR("socketcan_interface", "internal error: " << input_.id);
```
jschleicher, line 206: I'd suggest to print the output of `translateError()`, since it is more meaningful than interpreting the id as CAN frame:

```suggestion
                std::string error_description;
                translateError(input_.id, error_description);
                ROSCANOPEN_ERROR("socketcan_interface", "internal error: " << error_description);
```

Or maybe move this into CanLayer, where the output is duplicated...

jschleicher: On the other hand, the "controller problems" are output as device state afterwards in can_layer.h, so maybe both outputs in CanLayer and socketcan.h could be dropped...
```
[ INFO] [1562593491.507573769]: Current state: 2 device error: system:0 internal_error: 4 (controller problems;)
[ INFO] [1562593491.507604879]: Current state: 1 device error: system:0 internal_error: 4 (controller problems;)
```

mathias-luedtke:
@jschleicher: I have decided to not drop any log message (as part of this PR). Please test this final version again.

canopen_chain_node/src/ros_chain.cpp
mathias-luedtke: This is now tracked in #345

socketcan_interface/include/socketcan_interface/socketcan.h
jschleicher: The output with 5889ab66fa21a11f98cfc7f08451d7df4906b2d7 and this simple test case now reads:
```
[ERROR] [1564589661.190719770]: internal error: 4
[ INFO] [1564589661.190777365]: Current state: 2 device error: system:0 internal_error: 4 (controller problems;)
[ INFO] [1564589661.190813729]: Current state: 1 device error: system:0 internal_error: 4 (controller problems;)
[ERROR] [1564589661.191039114]: Received error frame: 20000004#0030000000000000
[ERROR] [1564589661.198288816]: CAN not ready
```

jschleicher:
@ipa-mdl Sorry for the delay! I've now tested 5889ab66fa21a11f98cfc7f08451d7df4906b2d7 and the output is the same as before. Should there be a change?

Otherwise I'm fine with this PR now.

mathias-luedtke:
> I've now tested 5889ab6 and the output is the same as before. Should there be a change?

Thanks!
The output should be the same, but the tests are passing again ;)

PR #290, opened 2018-06-13
fix bug with invalid update_ms param load

NeoPhix:
nh_priv_ -> sync_nh for right loading of update_ms interval

### DIFF
--- a/canopen_chain_node/src/ros_chain.cpp
+++ b/canopen_chain_node/src/ros_chain.cpp
@@ -284,7 +284,7 @@ bool RosChain::setup_sync(){
     }
 
     int update_ms = sync_ms;
-    if(sync_ms == 0) nh_priv_.getParam("update_ms", update_ms);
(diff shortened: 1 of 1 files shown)


mathias-luedtke (changes requested):
Thanks!

I am not sure if this is actually a bug in the code. I might have read it from the private namespace on purpose., because `update_ms` is used only if the sync is disabled.
This means that the documentation is wrong.

Since it was read from private namespace for years, this should not be changed.
As a compromise, it could be read from `sync/update_ms` first and from `~update_ms` as a fall-back.

mathias-luedtke:
@NeoPhix: Do you plan to work on this?

NeoPhix:
> @NeoPhix: Do you plan to work on this?

@ipa-mdl 
No, I don't. It was a side issue, which does not influence the required functional of the library. PR should be closed.

PR #294, opened 2018-07-10
added C++11 flag for canopen_motor_node

geoffviola:
I needed this change to compile for lunar on Ubuntu 16.04


mathias-luedtke:
This is not needed. (Travis passes for lunar.)
And it would result in unsolveable ABI breaks with other C++03 code.

I think this a side effect of bulding the `melodic-devel` of `ros_control`. (https://github.com/ros-controls/ros_control/pull/345)

PR #302, opened 2018-09-26
Add functions to disable specific error types in SocketCANInterface

redkite:
Similar to #264, but so far limited to the SocketCANInterface only.

Saves the state of the CAN error mask within the object and provides two functions to disable lost arbitration and controller problems. These two have been added explicitly in order to not disable other errors by accident. A private function within SocketCANInterface still provides an internal, generic interface.

Could also be added to the socketcan bridge. I cannot test it though, because we are using the SocketCANInterface only.

### DIFF
--- a/socketcan_interface/include/socketcan_interface/socketcan.h
+++ b/socketcan_interface/include/socketcan_interface/socketcan.h
@@ -199,11 +199,9 @@ protected:
             if(frame_.can_id & CAN_ERR_FLAG){ // error message
                 input_.id = frame_.can_id & CAN_EFF_MASK;
                 input_.is_error = 1;
-
                 LOG("error: " << input_.id);
-                setInternalError(input_.id);
-                setNotReady();
 
+                handleErrorFrame();
             }else{
                 input_.is_extended = (frame_.can_id & CAN_EFF_FLAG) ? 1 :0;
                 input_.id = frame_.can_id & (input_.is_extended ? CAN_EFF_MASK : CAN_SFF_MASK);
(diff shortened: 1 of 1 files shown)


mathias-luedtke (changes requested):
Thanks for your patch!
I am afraid, I cannot merge it (as-is).
It introduces ABI breaks and adds new interfaces in a sub class.

As I have commented before, I don't think that changing the error mask is the right way to solve this.
The error should get white-listed [here](https://github.com/ros-industrial/ros_canopen/blob/kinetic-devel/socketcan_interface/include/socketcan_interface/socketcan.h#L204) instead.

I could imagine a virtual, private function `handleErrorFrame` that  could get configured or overridden in sub classes.

redkite:
Sorry for the late reply, I was busy during the last week. I will have a look at possibilities to implement the changes and come back to you this week.

redkite:
@ipa-mdl Is this more along the lines what you would expect? One could then inherit from `SocketCANInterface` and replace the error handler.

I am still not entirely sure, if the error mask should be hidden in the `init` function. If I would like to add `CAN_ERR_BUSERROR` as a handled error, I still would have to replace the complete `init` function. Could you imagine merging a PR whith a virtual function `genErrorMask()` to work around such issues, or do you deem this completely unnecessary?

redkite:
@ipa-mdl Do you have any feedback on these changes?

redkite:
Closing this, no PR response for a year.

Timple:
Still a valid PR. I think the discussion is meanly happening here: https://github.com/ros-industrial/ros_canopen/pull/264

But so far it the conversation seems to come to halt every time :(. Feel free to chip in there!

Issue #267, opened 2018-02-07
socketcan_bridge: unrecoverable error "controller problems;, asio: Success"


WasabiFan:
I'm currently using `socketcan_bridge` to interface with CAN in my ROS project. While it does generally work, it will sometimes spit out a sequence of four lines like this:

```
[ERROR] [1517684645.914299608]: Error: controller problems;, asio: Success
```

After that, any further attempts to send a message result in:

```
[ERROR] [1517684649.054520857]: Failed to send message: 9e040005#010384fc7c.
```
I can send that same CAN frame from the command line. So, it's not an underlying driver failure, but it seems that the bridge doesn't recover from an intermittent bus error. Yes,  the ideal solution is to resolve whatever causes it, but if this ever happens in a production environment the whole system grinds to a halt.

This seems to be where the latter message is coming from:

https://github.com/ros-industrial/ros_canopen/blob/4b0345a568daf1417b31afa38c9c7fd1ac31245c/socketcan_bridge/src/topic_to_socketcan.cpp#L69-L73

I am not sure where the former originates or what produces it. In `dmesg`, I often see the following when, e.g., I plug in a USB device (my guess is that the USB processing is flooding interrupts or something of that nature):

```
[12842.644761] mttcan c310000.mttcan can0: mttcan_poll_ir: some msgs lost on in Q0
```

That may or may not be related.

I think #249 would fix the problem, but I don't think I am personally prepared to take on the fixes to that PR that were requested.

Could anyone suggest a fix or resolution?

mathias-luedtke:
> Could anyone suggest a fix or resolution?

Auto-recover would be an option, but some messages will get lost.
#244 needs to be fixed..

>  but if this ever happens in a production environment the whole system grinds to a halt.

If USB interrupts can break your application, I would not consider this a production environment.
A lowlatency kernel might help as a work-around.

PR #249, opened 2017-09-12
Adding simple autorecover - hopefully all of my build errors are fixed

watters-torc:
Maintainers: This is a feature we needed for a project, originally we were under the impression the socketcan_bridge and socketcan_interface were under BSD, it seems like this project (ros_canopen) has a bunch of mixed licenses. I don't necessarily see this as a production ready fix but it works for what we need. I'm able to make changes or take direction from the maintainers if you'd like work to further be done here.

Proposed fix / gap stop for #244

Issue #244, opened 2017-08-04
Add CAN error handling to socketcan_bridge


mathias-luedtke:
Currently, the socketcan_bridge nodes to not handle error.
There is no inteface to read the state or any way to recover the interface.

(via http://answers.ros.org/question/267980/socketcan_bridge_node-error-handling/)

windelbouwman:
Anyone wishing to reproduce this issue, can follow these steps:

Setup `vcan0`:

    $ sudo ip link add dev vcan0 type vcan
    $ sudo ip link set up vcan0

Run the socketcan bridge node:

    $ roscore
    $ rosrun socketcan_bridge socketcan_bridge_node _can_device:=vcan0

Send can frames, for example:

```python
import rospy
from can_msgs.msg import Frame

rospy.init_node('sender')
pub = rospy.Publisher('/sent_messages', Frame, queue_size=20)
rate = rospy.Rate(13)

counter = 1
while not rospy.is_shutdown():
    frame = Frame()
    frame.id = 13
    frame.dlc = 3
    frame.data = bytes([55, 11, counter])
    counter += 17
    while counter >= 256:
        counter -= 256
    pub.publish(frame)
    rate.sleep()
```

Inject an error with the below:

```python

import socket
import struct
import sys

assert sys.version_info.major == 3

CAN_ERR_FLAG = 0x2000_0000
CAN_ERR_LOSTARB = 0x2
CAN_ERR_BUSOFF = 0x40
CAN_ERR_MASK = 0x1FFF_FFFF


def pack_msg(msg_id, payload):
    """ Pack bytes according to socketcan. """
    ln = len(payload)
    flags = 0

    part1 = struct.pack('=IBB2x', msg_id, ln, flags)
    assert ln <= 8
    padding_size = 8 - ln
    part2 = payload + bytes([0] * padding_size)
    assert len(part2) == 8
    xmit = part1 + part2
    assert len(xmit) == 16
    return xmit


def emit(sock, msg_id, payload):
    xmit = pack_msg(msg_id, payload)
    sock.send(xmit)

def inject_error(sock, err):
    emit(sock, CAN_ERR_FLAG | err, bytes([0] * 8))

def main():
    device = 'vcan0'

    # Create socket:
    sock = socket.socket(socket.PF_CAN, socket.SOCK_RAW, socket.CAN_RAW)
    sock.bind((device,))

    try:
        emit(sock, 0x133, bytes([1,2,3]))
[...]

PR #310, opened 2018-11-05
If an error occurs try and recover from it if necessary

matt-attack:
We were having issues with the driver refusing to send packets after an error of some type occurred on the bus, likely due to the hardware on the other end. Simply recovering was able to fix the problem. This is probably not the best way to do it, but it definitely works.

Related to #267 

It may be good to just add a parameter to enable or disable this functionality and merge it in as a stopgap until the fancier solution #249 is complete. This node is useless to us without this functionality.


windelbouwman:
Related to #244

PR #318, opened 2019-02-11
FastDelegate using Variadic Templates and static_assert

JWhitleyWork:
This brings FastDelegate.h in socketcan_interface up to the now by using parameter packs instead of re-implementing the classes/functions for every possible number of parameters. It also cleans up that file a bit with regard to formatting and replaces all compile-time checks with static_assert (instead of crazy compiler hacks).

Continues the changes implemented in #282.


mathias-luedtke:
Interesting..
Are there any benifits at run-time?

JWhitleyWork:
Runtime performance should be the same. The compiler should generate almost identical code to what was in FastDelegate.h before and compilation will actually take slightly longer. However, development and API usage are simplified significantly.

JWhitleyWork:
Closing in favor of #320

Issue #103, opened 2015-07-07
Move socketcan_interface into dedicated repository


mathias-luedtke:
Might be reasonable for CAN-only development

JWhitleyWork:
Do you know if anyone is using socketcan_interface independently (anyone outside of this repository)?

Timple:
We do for some of our robots. For prototyping it is easier to use the can_msgs than to open up a socket connection.
But we don't mind it being in this repo actually.

mathias-luedtke:
For the released version it does not matter, but for developement t might be easier to just have the CAN-related packages (can_msgs, socketcan_interface, socketcan_bridge) without the ros_control dependencies.

samiamlabs:
I used socketcan_interface independently from this repo for a robot a couple of years ago. The hardware I was interfacing with was only partially using CANopen. Made a custom ros_control hardware interface.

I agree that having it in this repo is ok, but separating would be nice for development.

JWhitleyWork:
@ipa-mdl Can you create the new repo? Do we want to do this for just Melodic or back-port it for others?

mathias-luedtke:
> Can you create the new repo? 

Actually, I can't..

>  Do we want to do this for just Melodic or back-port it for others?

I would do it for the ROS2 port and ROS-N.
Not sure about melodic.

@gavanderhoorn: What do you think about splitting the repository?

gavanderhoorn:
@ipa-mdl wrote:
> @gavanderhoorn: What do you think about splitting the repository?

Is the desire mostly to be able to use `can_msgs` without any of the other packages, or really to have access to `socketcan_interface` only (with `can_msgs`)?

If the former: I'd take `can_msgs` out of this repository. If the latter: is it just about dependencies, or also about compilation times?

@ipa-mdl: what is the status of the packages in this repository? Mostly maintenance, or active development? Splitting repositories will complicate PR management as you have to sync everything to make sure they get merged in the proper order (for instance).

---

Edit: something else to think about: would you want to maintain commit history for `socketcan_interface`? If so: which approach would you like to take: mirror `ros_canopen` to new repository and delete everything except `can_msgs` and `socketcan_interface`, or do a `git filter-branch`?

mathias-luedtke:
I would keep `can_msgs`, `socketcan_interface` and `socketcan_bridge` together. to split into CAN and CANopen set.

> If the latter: is it just about dependencies, or also about compilation times?

[...]

PR #320, opened 2019-02-12
Replacing FastDelegate with std::function and std::bind.

JWhitleyWork:
FastDelegate is not optimized for usability with C++11/14 features. In addition, the std library functions are much more portable, if possibly slightly slower.

I attempted to maintain API compatibility but the function signatures are so different between FastDelegate and std::function/std::bind that I ended up creating a utility function with FastDelegate's signature. I thought about inheriting from std::function to replace the typedefs but I couldn't think of a way to do it and have a constructor with the same signature of FastDelegate without knowing/reimplementing some of the std::function internals. Please let me know if you can think of a way that will maintain API compatibility.

This is an alternative to PR #318.


JWhitleyWork:
@ipa-mdl - I found a way to not break API! Hooray! It seems a little hack-y, though. Can you review?

mathias-luedtke:
Overall: :+1: 

Two comments/ideas:
* I'd like to allow passing just `std::function`, so we could just add a *DelegateFunc type for use in interfaces
* Is it possible to write a template that provides the *Delegate wrapper?

The first one should be easy to do.
The second should be possible as well, if not a macro could be an alternative.
I will try to come up with something.

What do you think?

JWhitleyWork:
@ipa-mdl - Since the *Delegate classes inherit from `std::function`, they have all of it's default constructors (including the copy and move constructors) so you can do something like:

    FrameDelegate fd = std::bind(...)

or

    FrameDelegate fd(std::bind(...))

As well as the FastDelegate-like forms for which I created new constructors. I think this covers your first requirement. I'm not sure I understand your second requirement. Could you elaborate a little bit more or provide a use case?

mathias-luedtke:
> Since the *Delegate classes inherit from `std::function`, they have all of it's default constructors (including the copy and move constructors)

This should not be the case (?).
But you can pass *Delegate instances to the `std::function` constructors 

> so you can do something like:

Yep, you are right, this should work :)
I thought (but did not write..) that the *Delegate classes could get deprecated at some point.

>  Could you elaborate a little bit more or provide a use case?

I'd like to have something like this:

```c++
using FrameFunc = std::function(void(const Frame&)>)
using FrameDelegate = DelegateHelper<FrameFunc>;
```

or

```c++
using FrameDelegate = DelegateHelper<void(const Frame&)>;
using FrameFunc = FrameDelegate::func_type;
```

I tried to implement `DelegateHelper` yesterday, but i did not yet find a solution that works for an arbitrary number( we need 0-2 )of arguments / placeholders..

JWhitleyWork:
So, just to clarify, you want FrameDelegate to essentially be a specialization of `std::function` that specifically uses the function signature `void(const Frame&)`, correct? What are you imagining the constructor for a FrameDelegate would look like? I did this before I implemented the "non-api-breaking" version of this but it required a call to `std::bind` in order to create a delegate for the specific function that you wanted attached to the function. Below are a few options of how this could look, both with and without using `std::bind`. Let me know which one you want and I'll get that in.

1.

    FrameDelegate foo1(std::bind(&CommInterfaceChild::func, &instance, std::placeholders::_1);
    TimerDelegate foo2(std::bind(&TimerChild::func, &instance);
    ReadDelegate foo3(std::bind(&ObjectStorageChild::func, &instance, std::placeholders::_1, std::placeholders::_2);

2.

    FrameDelegate foo1 = DelegateHelper<FrameDelegate>(&CommInterfaceChild::func, &instance);
    TimerDelegate foo2 = DelegateHelper<TimerDelegate>(&TimerChild::func, &instance);
    ReadDelegate foo3 = DelegateHelper<ReadDelegate>(&ObjectStorageChild::func, &instance);

3.

    using FrameDelegate = DelegateHelper<void(const Frame&)>;
    using TimerDelegate = DelegateHelper<bool(void)>;
    using ReadDelegate = DelegateHelper<void(const ObjectDict::Entry&, String&)>;

    FrameDelegate foo1(&CommInterfaceChild::func, &instance);
    TimerDelegate foo2(&TimerChild::func, &instance);
    ReadDelegate foo3(&ObjectStorageChild::func, &instance);

FYI - The current implementation is similar to option 3 but the order of the arguments is reversed.

Also, do you care about breaking API? If not, I'll implement this and change all references to it as well.

mathias-luedtke:
I'd like to make the `std:function` interface the default, which will be used together with `std::bind` or lambda functions (option 1).
For backwards compatibility, it would be nice to have an API-compatible *Delegate type available (option 3), which could be marked as deprecated (later).

mathias-luedtke:
@JWhitleyAStuff: I am preparing the *Func patch right now and open a PR to your fork soon.

mathias-luedtke:
I have now figured out how to implement `DelegateHelper`:

```c++
template <typename T> class DelegateHelper : public T {
public:
    template <typename Object, typename Instance, typename ...Args>
    DelegateHelper(Object &&o, typename T::result_type (Instance::*member)(Args... args)) :
        T([&o, fn=std::mem_fn(member)](Args... args) -> typename T::result_type { return fn(o, args...); })
    {
    }
    template <typename Callable>
    DelegateHelper(Callable &&c) : T(c)
    {
    }
};

using StateFunc = std::function<void(const State&)>;
using StateDelegate = DelegateHelper<StateFunc>;
```

Perhaps `std::forward`  could be added in some places ^^

JWhitleyWork:
Do you want me to add this to this PR?

mathias-luedtke:
> Do you want me to add this to this PR?

Yes, please try it.
It would be good as well to have a small test that tests the Delegate to work as before.

I will try to provide a benchmark.

JWhitleyWork:
@ipa-mdl - In the suggested function above, you're using an initialized lamda capture. This requires C++14. Should we reorganize for C++11 or should we add -std=C++14 to the CMakeLists.txt?

JWhitleyWork:
I have added the code and made the necessary changes. So that there is easier access to the DelegateHelper without all of SocketCANInterface, I have moved it to socketcan_interface/helpers.h. Please let me know if this is OK. You may have to develop the test as I'm not as familiar with the concepts behind how the Delegates were used.

mathias-luedtke:
> This requires C++14. Should we reorganize for C++11 or should we add -std=C++14 to the CMakeLists.txt?

No, C++14 is the default for Ubuntu 18.04 and melodic.
I was not able to do this in C++11, but we can reorganize if possible.

JWhitleyWork:
If this is only targeting Melodic and above (and eventually ROS2), then I'm fine with just leaving it. My goal behind this was to get the whole repo to a ROS2-ready state anyway. Everything compiles and all tests succeed (minus the one that I hope you'll write for Delegates). If you're happy with the changes, I'm good for merge minus that last test.

FYI - I'm working on the ROS2 version here: https://github.com/JWhitleyAStuff/ros_canopen/tree/crystal-devel

mathias-luedtke:
I have now added some API tests.
Next step: simple benchmarking tool and a CANopen hardware test.

> FYI - I'm working on the ROS2 version here: https://github.com/JWhitleyAStuff/ros_canopen/tree/crystal-devel

Great!
Not sure if we should split the repositories to simplify the migration (#103).

mathias-luedtke:
I have run some benchmark and the new version is around 30% slower, but still at very high level (~10Mhz if I recall it correctly).
Combined with the dispatcher and the actual CAN device, this is not noticeable anymore.

However, somehow my tests crash in Docker (or at least don't output results).
I am still investigating..

JWhitleyWork:
@ipa-mdl - Since we were already working on socketcan_interface, I went ahead and replaced all typedefs with using aliases. They're functionally identical except that aliases can be used in templated functions and typedefs can't. Figured it wouldn't hurt and it's helpful for the ROS2 transition anyway.

JWhitleyWork:
@ipa-mdl Ping

mathias-luedtke:
Sry, it took me time to hunt the segmentation fault.
https://github.com/JWhitleyAStuff/ros_canopen/pull/5 should fix it.
After the merge  the branch needs to be rebased (proper commits without merge) and then we need to test it with real HW.

mathias-luedtke:
@JWhitleyAStuff: Do you have to rebase the branch properly or should I got ahead a push to you branch?
The tests can be fixed by dropping ef420b797d42e0e8737ad795f68919fc844f43f2 during the rebase.

JWhitleyWork:
@ipa-mdl Just rebased on `melodic-devel`.

JWhitleyWork:
@ipa-mdl - Isn't ef420b7 necessary to run the test?

mathias-luedtke:
> Isn't ef420b7 necessary to run the test? 

No, it was just a standalone node. I refactored it later into a test.
I have now removed this commit and deprecated the delegates (https://github.com/ipa-mdl/ros_canopen/commits/JWhitleyAStuff-maint/std_function).
I went ahed and rebased it as well: https://github.com/ipa-mdl/ros_canopen/commits/JWhitleyAStuff-maint/std_function-rebased
@JWhitleyAStuff: If this rebased version is okay for you, I would force-push it onto your branch.

I already tested the version with a CANopen manipulator and it seems to work :)
@jschleicher: It would be great, if you could give it a try as well.

JWhitleyWork:
@ipa-mdl - Go ahead and force-push it. The benchmark looks fine.

mathias-luedtke (approved):
@JWhitleyAStuff: Thanks a lot for these many iterations! It is ready for merge.
I am just waiting for @jschleicher's feedback.

jschleicher (approved):
Works as before; no changes detected.

mathias-luedtke:
Little glitch fixed in 9a393ae735306f077037145c70cd99df31c3223c. ROS buildfarm complained about the "deprecated" warning in the test of the backwards compatibilty. ..

PR #331, opened 2019-05-09
Add no_mode_state parameter

mistoll:
After a firmware upgrade our motor (Nanotec PD4-C) can't switch to "Operation Enabled" without a mode selected any more.

Similar to "switching_state" parameter a "no_mode_state" parameter was added. That parameter allows us to set the state when no mode is selected.

The default behaviour should be the same as before. This code works for us, however there might be something missing. Reviews welcome.

### DIFF
--- a/canopen_402/src/motor.cpp
+++ b/canopen_402/src/motor.cpp
@@ -116,40 +116,11 @@ Command402::TransitionTable::TransitionTable(){
     // fault reset
     /*15*/ add(s::Fault, s::Switch_On_Disabled, Op((1<<CW_Fault_Reset), 0));
 }
-State402::InternalState Command402::nextStateForEnabling(State402::InternalState state){
-    switch(state){
-    case State402::Start:
-        return State402::Not_Ready_To_Switch_On;
 
-    case State402::Fault:
-    case State402::Not_Ready_To_Switch_On:
-        return State402::Switch_On_Disabled;
-
-    case State402::Switch_On_Disabled:
-        return State402::Ready_To_Switch_On;
-
-    case State402::Ready_To_Switch_On:
-        return State402::Switched_On;
-
-    case State402::Switched_On:
-    case State402::Quick_Stop_Active:
-    case State402::Operation_Enable:
-        return State402::Operation_Enable;
-
-    case State402::Fault_Reaction_Active:
-        return State402::Fault;
-    }
-}
-
-bool Command402::setTransition(uint16_t &cw, const State402::InternalState &from, const State402::InternalState &to, State402::InternalState *next){
+bool Command402::setTransition(uint16_t &cw, const State402::InternalState &from, const State402::InternalState &to){
     try{
         if(from != to){
-            State402::InternalState hop = to;
-            if(next){
-                if(to == State402::Operation_Enable) hop = nextStateForEnabling(from);
-                *next = hop;
-            }
-            transitions_.get(from, hop)(cw);
+            transitions_.get(from, to)(cw);
         }
         return true;
     }
@@ -290,6 +261,8 @@ ModeSharedPtr Motor402::allocMode(uint16_t mode){
 bool Motor402::switchMode(LayerStatus &status, uint16_t mode) {
 
     if(mode == MotorBase::No_Mode){
+        if(!switchState(status, no_mode_state_)) return false;
+
         boost::mutex::scoped_lock lock(mode_mutex_);
         selected_mode_.reset();
         try{ // try to set mode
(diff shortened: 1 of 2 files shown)


canopen_402/src/motor.cpp
```diff
@@ -481,7 +481,7 @@ void Motor402::handleInit(LayerStatus &status){
         control_word_ = 0;
         start_fault_reset_ = true;
     }
-    if(!switchState(status, State402::Operation_Enable)){
+    if(!switchState(status, no_mode_state_)){
```
mathias-luedtke: This should only be overwritten if no mode was selected.

canopen_402/src/motor.cpp
```diff
@@ -511,6 +511,7 @@ void Motor402::handleInit(LayerStatus &status){
     switchMode(status, No_Mode);
 }
 void Motor402::handleShutdown(LayerStatus &status){
+    switchState(status, no_mode_state_);
```
mathias-luedtke: In worst case, this would try to switch on the motor first.
I would move this into  `switchMode`.

mathias-luedtke (changes requested):
The introduction of the option is reasonable.
However, not all side-effect are covered.

At the moment the driver only sends the control word if operation is enabled.

canopen_402/src/motor.cpp
mistoll: It's hard for me to anticipate the purpose of that line. Is Operation_Enabled required for the Homing mode? Or does is support the case where Homing is disabled? To be clear: switchMode does switch to Operation_Enabled anyway.
A propsal:
```
// if(!switchState(status, no_mode_state_)){
//   status.error("Could not enable motor");
//   return;
//}

ModeSharedPtr m = allocMode(MotorBase::Homing);
if(!m){
    if(!switchMode(status, No_Mode)){
        status.error("Could not enable motor");
    }
    return; // homing not supported
}
```
(with no_mode_state handled in switchMode)

mathias-luedtke: > with no_mode_state handled in switchMode

:+1: 

But I think no_mode needs to be handled in more places, e.g., https://github.com/ros-industrial/ros_canopen/blob/162c0a8b20fda2f2a4b13ab1c26e1437beeacdb3/canopen_402/src/motor.cpp#L417

mistoll: Is that line required at all? It prevents the operation mode specific bits from being set when state is not Operation_Enabled. But a check if a mode is selected is already in place
https://github.com/ros-industrial/ros_canopen/blob/162c0a8b20fda2f2a4b13ab1c26e1437beeacdb3/canopen_402/src/motor.cpp#L421

In no mode case that won't do anything no matter what state we're in. Or am I missing something?

mistoll: Please have a look at 1a82bcc.
If you don't like what I did to setTransition, the same thing would be possible with the old signature, but I think it's cleaner this way.
If you think handling no_mode_state in handleWrite and/or handleHalt is required, I can add that too.

canopen_402/src/motor.cpp
```diff
@@ -352,22 +325,46 @@ bool Motor402::switchMode(LayerStatus &status, uint16_t mode) {
...
     time_point abstime = get_abs_time(state_switch_timeout_);
     State402::InternalState state = state_handler_.getState();
     target_state_ = target;
     while(state != target_state_){
         boost::mutex::scoped_lock lock(cw_mutex_);
-        State402::InternalState next = State402::Unknown;
-        if(!Command402::setTransition(control_word_ ,state, target_state_ , &next)){
+        State402::InternalState hop = target_state_;
+        if(target_state_ == State402::Operation_Enable || target_state_ == no_mode_state_)
+            hop = nextStateForEnabling(state);
```
mathias-luedtke: This might not be possible for all cases..

canopen_402/src/motor.cpp
mathias-luedtke: I guess, we need a mechanism like `nextStateForEnabling`  for all possible options for `no_mode_state` (or all).

mathias-luedtke: Which state is okay for Nanotec? `Switched_On`?
This might even be a senseful default choice..

canopen_402/src/motor.cpp
mistoll: Let's have a look at the power state machine.

IMO the following states are possible "no mode states":

1. Switch on disabled
2. Ready to switch on
3. Switched on
4. Operated enabled

Should we check that and add an exception?

If switching upwards (towards Operation enabled) then "nextStateForEnabling" should return the correct next state. It's just like switching to Operation enabled but will succeed earlier.

However when switching downwards then "nextStateForEnabling" is not correct. If not in fault or quick stop we can switch to the downward state directly. If in fault or quick stop "nextStateForEnabling" should work again.

The difficulty here is to determine when to use nextStateForEnabling. IMO this information is present in the transistion table. So I'd use nextStateForEnabling whenever setTransistion returns false. Let me add that ...

canopen_402/src/motor.cpp
mistoll: > Which state is okay for Nanotec?

From [Nanotec PD4-C documentation](https://en.nanotec.com/products/manual/PD4C_CAN_EN/general%252Fds402_power_state_machine.html/):

- Switch on disabled
- Ready to switch on
- Switched on

`switched_on` would be a good default choice! Can you verify if that works with other motors, too?

mathias-luedtke:
This needs a thorough review and probably some refacturing

mistoll:
We'v been working with this branch for a while and didn't have any problems related to this patch.
Is there a chance, this gets merged soon?

PR #334, opened 2019-06-11
reset node on initialization

jschleicher:
Resetting the communication only loads default values for the pdo mapping in the motor firmware according to CiA 301. I suggest with this PR to also reset the application registers to get a roslaunch behaviour that is similar to a power-on-reset of the motor (firmware) to fully recover from errors and values from the previous run.

What do you think?

### DIFF
--- a/canopen_master/src/node.cpp
+++ b/canopen_master/src/node.cpp
@@ -178,7 +178,7 @@ void Node::handleInit(LayerStatus &status){
 
     sdo_.init();
     try{
-        if(!reset_com()) BOOST_THROW_EXCEPTION( TimeoutException("reset_timeout") );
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
In general: :+1: 
Unfortunately, some devices are not compatible with this change.
E.g., Elmo devices seem to connect the CAN lines to ground during `reset()`, which breaks the CAN communication in a chain.

I am afraid, we have to make this somehow configurable.

mathias-luedtke:
> I am afraid, we have to make this somehow configurable.

What do you think of a `deep_reset` argument in an additional constructor (ABI..) that gets passed down from `canopen_chain_node`?

jschleicher:
Sounds good. Probably this is better than an additional `/reset` service that the user can trigger in addtion before the `/init` service.

Issue #319, opened 2019-02-12
Log error message, if initialization / recover fails


JonathanGruner:
When calling the service interface to init / recover the motor, there is an error string in the response of the service call, if something fails. Calling the service from the launch file on the other hand does not give any error indication to the user.

There should be the same error string printed to the console:

```
[ INFO] [1549975063.533360109]: Initializing XXX
[ INFO] [1549975063.533741997]: Current state: 1 device error: system:0 internal_error: 0 (OK)
[ INFO] [1549975063.534281762]: Current state: 2 device error: system:0 internal_error: 0 (OK)
--> Error, but no warning / error message
```

mathias-luedtke:
Thanks for reporting!
It should be quite easy to add this in `RosChain::handle_init` & Co.
I could imaging some scoped response logger.

jschleicher:
@ipa-mdl
Following the discussion in https://github.com/ros-industrial/ros_canopen/pull/286#discussion_r299515309, a plain `ROS_WARN_*` output in https://github.com/ros-industrial/ros_canopen/blob/6890527ee9af45afd8c607ff86b92e699de79fd4/canopen_chain_node/src/ros_chain.cpp#L108-L109 would be alright, correct? What is the diagnostics logger in canopen_chain_node about?

mathias-luedtke:
> a plain ROS_WARN

`ROS_WARN` or `ROS_ERROR`.

> What is the diagnostics logger in canopen_chain_node about?

In general?

PR #343, opened 2019-07-08
Log output of /init and /recover services

jschleicher:
Duplicate the reason that is put into the service response string onto
the terminal in case of failures.
Implements #319


mathias-luedtke (changes requested):
This does not print a warning in all cases (exceptions).
I wil try to come up with a proposal.

jschleicher:
@ipa-mdl How about outputting the content of `res.message` at the end of the function (see my second commit)?

mathias-luedtke:
It still does not print the warning in all case.
(Mostly, because I like my code to jump out as soon as possible..)

> I wil try to come up with a proposal.

Hopefully next week..

PR #349, opened 2019-08-08
added support for SYNC counter in SimpleSyncLayer

### LINKED ISSUE #348: SimpleSyncLayer does not handle overflow sync counter
Setting the overflow param to any value between 0 and 240 should result into a counter value inside the sync frame. The external sync (bcm_sync) is doing it correctly but the internal sync (SimpleSyncLayer) is not.

relevant code section: https://github.com/ros-industrial/ros_canopen/blob/melodic-devel/canopen_master/src/master_plugin.cpp#L53

mathias-luedtke:
fixes #348 

@benmaidel: please test

### DIFF
--- a/canopen_master/src/master_plugin.cpp
+++ b/canopen_master/src/master_plugin.cpp
@@ -43,6 +43,21 @@ public:
 
 class SimpleSyncLayer: public ManagingSyncLayer {
     time_point read_time_, write_time_;
+    can::Frame frame_;
+    uint8_t overflow_;
+
+    void resetCounter(){
+        frame_.data[0] = 1; // SYNC counter starts at 1
+    }
+    void tryUpdateCounter(){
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
Not sure, if the counter should get reset at some point.

benmaidel (approved):
Successfully tested with cob base. Sync counter is counting if enabled.

mathias-luedtke:
@benmaidel: I guess you need this for kinetic as well, don't you?

benmaidel:
> @benmaidel: I guess you need this for kinetic as well, don't you?

right, for now I've cherry picked the commit into kinetic-devel. Works fine! Now I've also seen your backport to kinetic https://github.com/ros-industrial/ros_canopen/pull/351 :+1: 
@ipa-mdl thx for support

PR #356, opened 2019-09-30
Log services

### LINKED ISSUE #319: Log error message, if initialization / recover fails

mathias-luedtke:
closes #319, replaces #343 

@jschleicher: please give this a try

### DIFF
--- a/canopen_chain_node/src/ros_chain.cpp
+++ b/canopen_chain_node/src/ros_chain.cpp
@@ -90,8 +90,48 @@ void RosChain::run(){
     }
 }
 
+template <typename T> class ResponseLogger {
+protected:
+    bool logged;
+    const T& res;
+    const std::string command;
+public:
+    ResponseLogger(const T& res, const std::string &command) : logged(false), res(res), command(command){}
+    ~ResponseLogger() {
+        if(!logged && !res.success){
+            if (res.message.empty()){
+                ROS_ERROR_STREAM(command << " failed");
+            }else{
+                ROS_ERROR_STREAM(command << " failed: " << res.message);
+            }
+            logged = true;
+        }
+    }
+};
+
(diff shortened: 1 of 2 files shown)


canopen_chain_node/src/ros_chain.cpp
```diff
@@ -90,8 +90,48 @@ void RosChain::run(){
...
+    const T& res;
+    const std::string command;
+public:
+    ResponseLogger(const T& res, const std::string &command) : logged(false), res(res), command(command){}
+    ~ResponseLogger() {
+        if(!logged && !res.success){
+            if (res.message.empty()){
+                ROS_ERROR_STREAM(command << " failed");
+            }else{
+                ROS_ERROR_STREAM(command << " failed:" << res.message);
```
jschleicher:
```suggestion
                ROS_ERROR_STREAM(command << " failed: " << res.message);
```

jschleicher (changes requested):
Yep, works well.

Just a tiny nit-picking comment.

mathias-luedtke:
Thanks!

I will rebase it later and merge.

jschleicher (approved):

mathias-luedtke:
This introduced [a regression](https://travis-ci.org/ros-industrial/ros_canopen/jobs/592239901) for Debian stretch, mostly because of an outdated version of `console_bridge`, which defines a `logWarn` macro.

PR #359, opened 2019-10-02
Second Try at Dashing Port

JWhitleyWork:
Re-opening this PR with a cleaned-up `git` history.

All packages have been ported with the exception of `canopen_motor_node` to ROS Dashing.

The following are issues or have yet to be addressed:

1. Some tests on `socketcan_interface` are failing.
2. Some tests on `socketcan_bridge` are disabled and need to be converted to use [launch_testing](https://index.ros.org/p/launch_testing/).
3. Nodes here should really be [LifeCycle Nodes](https://github.com/ros2/demos/blob/master/lifecycle/README.rst) since many of them are likely to be used in system-critical roles.
4. Certain nodes should be made available as [Components](https://index.ros.org/doc/ros2/Tutorials/Composition/).


JWhitleyWork:
Thanks to @ipa-mdl and @samiamlabs for your contributions!

JWhitleyWork:
@ipa-mdl - Is it possible to merge this port as-is and we continue to work on it on the `dashing-devel` branch?

JWhitleyWork:
@ipa-mdl Bump

mathias-luedtke:
@JWhitleyAStuff: Sry, I did not have much time to deal with this PR.

> Is it possible to merge this port as-is and we continue to work on it on the dashing-devel branch?

I don't want to push this to devel. It is not in a proper shape (for release, etc.) yet.
If you really want to get it out of your fork, I could push it to `dashing-experimental` branch.
In constrast to `dashing-devel`, we can rebase it etc.

I'd like to start with merging style changes first (incl h->hpp).
Then the non-ROS packages (catkin->ament) and finally the ROS packages (ROS1->ROS2).
The latter needs more attention, because I not super familiar with ROS2 yet and I want to take the chance to refactor it a little bit.

Long story short:
- push to `dashing-eperimental` (no maintenance, just for refactoring)
- create feature branches and merge them into `dashing-devel`
- Some changes should get backported to `melodic`

JWhitleyWork:
@ipa-mdl I'm OK with this. With regard to getting the release process going, I am mostly concerned with getting `can_msgs` ported and released since many of AutonomouStuff's drivers rely on `can_msgs/Frame` (and I believe several others are in the same situation). Your plan is OK with me for the overall porting process, though.

mathias-luedtke:
> I am mostly concerned with getting can_msgs ported and released

That should be rather easy :)
I just have to check if I can keep the other catkin packages in the repository during our refactoring.
Another option would be to use https://github.com/ros-industrial/socketcan_interface for ROS2.
I will try to port at least `can_msgs` as soon as possible.

socketcan_interface/CMakeLists.txt
```diff
@@ -1,151 +1,132 @@
...
+  ${pluginlib_INCLUDE_DIRS}
   ${Boost_INCLUDE_DIRS}
-  ${catkin_INCLUDE_DIRS}
-  ${console_bridge_INCLUDE_DIRS}
 )
 
 # ${PROJECT_NAME}_string
-add_library(${PROJECT_NAME}_string
-    src/string.cpp
+add_library(${PROJECT_NAME}_string SHARED
```
mathias-luedtke, line 25: does this really have to be shared?

socketcan_interface/CMakeLists.txt
```diff
@@ -1,151 +1,132 @@
...
-  ${console_bridge_LIBRARIES}
-  ${catkin_LIBRARIES}
+  ${pluginlib_LIBRARIES}
   ${Boost_LIBRARIES}
 )
 
+pluginlib_export_plugin_description_file(${PROJECT_NAME} socketcan_interface_plugin.xml)
+
+ament_export_dependencies(
+  ament_cmake
```
mathias-luedtke: why does `ament_cmake` need to be exported?

socketcan_interface/package.xml
```diff
@@ -13,16 +13,17 @@
   <url type="repository">https://github.com/ros-industrial/ros_canopen</url>
   <url type="bugtracker">https://github.com/ros-industrial/ros_canopen/issues</url>
 
-  <buildtool_depend>catkin</buildtool_depend>
+  <buildtool_depend>ament_cmake_auto</buildtool_depend>
 
+  <depend>pluginlib</depend> 
   <depend>boost</depend>
-  <depend>class_loader</depend>
-  <depend>libconsole-bridge-dev</depend>
```
mathias-luedtke, line 20: why did you drop  `libconsole-bridge-dev`?

socketcan_interface/CMakeLists.txt
JWhitleyWork: Since it is used in builds further down which are then used themselves in further builds, it avoids the linker thinking there are multiple definitions of the functions included in this library. I originally tried doing this without it being SHARED but got multiple linker errors because of this.

socketcan_interface/package.xml
JWhitleyWork: I'm not sure why I dropped it. Re-added.

socketcan_interface/CMakeLists.txt
JWhitleyWork: No. Removed.

socketcan_interface/CMakeLists.txt
mathias-luedtke: It *should* work with static as well, but I will have a look later.

PR #363, opened 2019-11-04
port can_msgs to ament build system

mathias-luedtke:
extracted from #359 

I am not sure, if we should add `AMENT_IGNORE` files and remove them later.

(@JWhitleyAStuff )

### DIFF
--- a/can_msgs/package.xml
+++ b/can_msgs/package.xml
@@ -1,23 +1,31 @@
 <?xml version="1.0"?>
-<package format="2">.
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>can_msgs</name>
   <version>0.8.2</version>
   <description>CAN related message types.</description>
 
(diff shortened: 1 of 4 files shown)


JWhitleyWork:
Did you intend to target this PR to the `melodic-devel` branch?

I would suggest using `COLCON_IGNORE` files.

can_msgs/package.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<package format="2">.
+<package format="3">.
```
JWhitleyWork: After the `<?xml version="1.0"?>`, this should have `<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>`. `ament_xmllint` can then use the XSD to validate the contents of this file.

mathias-luedtke:
> Did you intend to target this PR to the melodic-devel branch?

No, I will retarget it :)

JWhitleyWork (approved):
LGTM.

mathias-luedtke:
released: https://github.com/ros/rosdistro/pull/22878

JWhitleyWork:
Awesome! Thanks. Now on to everything else!

PR #371, opened 2019-12-17
configure GitHub action

mathias-luedtke:


gavanderhoorn:
@ipa-mdl: it seems this PR didn't remove the Travis config. Was that on purpose?

mathias-luedtke:
@gavanderhoorn: Yes and no.. I did not see a reason to remove them and wanted to have a fallback. This way I could test things as well in my fork. And it is another test case for industrial_ci.

PR #388, opened 2020-04-16
Option to ignore error frames

yanick-douven:
Finalized implementation of #362 with the addition of:
- ROS parameter parsing (ignore-errors/error-name=true/false)
- `CAN_ERR_LOSTARB` is ignored by default
- tested on physical CAN bus


Timple:
Since technically speaking the default behavior changes with this PR. Could this PR be evaluated before the official noetic release?

To prevent any discussions later if defaults can/might/should be changed within a release.

mathias-luedtke:
> Could this PR be evaluated before the official noetic release?

I had a brief look already.
I cannot promise that I will have enough time until May 23.

My quick review:
- the default error mask should be kept as-is, perhaps configurable
- the *handling* of the errors should be configurable, arbitration lost error should get ignored
- `serializable_map` is not needed, `can::Settings` should be subclassed if you want to offer an API
- https://github.com/ros-industrial/ros_canopen/blob/162c0a8b20fda2f2a4b13ab1c26e1437beeacdb3/canopen_motor_node/src/motor_chain.cpp#L8 should be used for ROS, perhaps templated to not depend on `XmlRpcValue` directly
- `canopen_chain_node` should pass the settings to `CANLayer` somehow.

> To prevent any discussions later if defaults can/might/should be changed within a release.

I don't want to branch off for noetic.
Only the arbitration lost error should get handled differently.
This change will affect melodic and noetic.

yanick-douven:
@ipa-mdl Thank you for taking the time to review! Few comments/considerations:

> * the default error mask should be kept as-is, perhaps configurable
> * the _handling_ of the errors should be configurable, arbitration lost error should get ignored

This means a possible high load from kernel to user space, as mentioned [here](https://github.com/ros-industrial/ros_canopen/pull/264#issuecomment-545515544). Why should the kernel still report all errors if we are ignoring them anyways in the user space?

> * `serializable_map` is not needed, `can::Settings` should be subclassed if you want to offer an API

I will move (parts of) `serializable_map` into a subclassed version of `can::Settings`.

> * https://github.com/ros-industrial/ros_canopen/blob/162c0a8b20fda2f2a4b13ab1c26e1437beeacdb3/canopen_motor_node/src/motor_chain.cpp#L8
>    should be used for ROS, perhaps templated to not depend on `XmlRpcValue` directly

After sending the settings to `can::SocketCANInterface::init()` we are leaving the ROS side of things, so what would we need `XmlRpcSettings` for? These `ignored_errors`-settings are generic and not ROS-specific.

> * `canopen_chain_node` should pass the settings to `CANLayer` somehow.

I don't understand exactly what you mean, does this refer to the handling of errors?

> This change will affect melodic and noetic.

Good to hear!

mathias-luedtke:
superseded by https://github.com/ros-industrial/ros_canopen/pull/362

Issue #399, opened 2020-07-31
catkin_make error when using socketcan_interface to build robotiq package


xianyitiantim:
Hi, I am trying to attach a robotiq 2-finger gripper on the ur5 arm robot in ROS. I downloaded the package from https://github.com/ros-industrial/robotiq and I am using ubuntu 18.04 and ROS melodic version. To bulid robotiq package downloaded from the link mentioned, we would need socketcan_interface package. However, when I tried to catkin_make this package in my workspace, it seems that there is one node that could not be bulit, and the error is as follows:

`CMakeFiles/robotiq_3f_gripper_can_node.dir/src/robotiq_3f_gripper_can_node.cpp.o: In function `boost::date_time::counted_time_system<boost::date_time::counted_time_rep<boost::posix_time::millisec_posix_time_system_config> >::subtract_times(boost::date_time::counted_time_rep<boost::posix_time::millisec_posix_time_system_config> const&, boost::date_time::counted_time_rep<boost::posix_time::millisec_posix_time_system_config> const&)':
/home/xian/ros_ws/src/ros_canopen-melodic-devel/socketcan_interface/include/socketcan_interface/socketcan.h:24: multiple definition of `can::parse_error_mask(std::shared_ptr<can::Settings const>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned int)'
CMakeFiles/robotiq_3f_gripper_can_node.dir/src/robotiq_3f_gripper_control/robotiq_3f_gripper_can_client.cpp.o:/home/xian/ros_ws/src/ros_canopen-melodic-devel/socketcan_interface/include/socketcan_interface/socketcan.h:24: first defined here
collect2: error: ld returned 1 exit status
robotiq/robotiq_3f_gripper_control/CMakeFiles/robotiq_3f_gripper_can_node.dir/build.make:152: recipe for target '/home/xian/ros_ws/devel/lib/robotiq_3f_gripper_control/robotiq_3f_gripper_can_node' failed
make[2]: *** [/home/xian/ros_ws/devel/lib/robotiq_3f_gripper_control/robotiq_3f_gripper_can_node] Error 1
CMakeFiles/Makefile2:2471: recipe for target 'robotiq/robotiq_3f_gripper_control/CMakeFiles/robotiq_3f_gripper_can_node.dir/all' failed
[...]

PR #400, opened 2020-07-31
make parse_error_mask a static member function

mathias-luedtke:
this prevents from duplicated definitions (like #399 )

@timple: please review


Timple (approved):
Neat solution

PR #402, opened 2020-08-22
test for foxy and rolling as well

mathias-luedtke:


mathias-luedtke:
Done for GitHub actions in #417

Issue #268, opened 2018-02-14
Add support for boolean datatype


mathias-luedtke:
Support for boolean data could be added for SDOs.
However, the PDO granularity will be kept at 8, so single bits cannot be mapped for performance reasons.

The following should be suffiicient to get basic support:
1. add data type enum
2. add parser
3. add C type to lookup, `bool`?
4. prevent mapping of objects with type 1
5. testing

It should more or less straight-forward, but needs some testing with devices that offer bit-access.

rostest:
To support further can devices and eds-files for a customer, the boolean datatype is required soon. An implementation of 1-bit SDOs would be extremly helpfull. Would it be possible in the near future?

> It should more or less straight-forward, but needs some testing with devices that offer bit-access.

I have access to such a device and would like to help. If f.e. some beta-version is available for (pre-)testing, please give me a hint. Many thanks and best regards!

mathias-luedtke:
>  Would it be possible in the near future?

Yes, if someone implements it. I don't have time (and no device) to do it.
That's why I added the "help wanted" label..

lkm1321:
Hi, I actually just added support for booleans in my fork, because my EDS file wasn't parsing because of boolean. I don't have the device in question with me at the moment, but I verified that:
1. EDS file containing booleans are parsed without throwing a std::bad_cast
2. When the user requests the object to be published, it's registered as an std_msgs/Bool topic

I'm happy to create a PR. Does it add on to melodic or melodic-devel?

PR #403, opened 2020-08-31
Support Boolean variables

lkm1321:
Hi, 

This PR adds support for Boolean (DataType=0x0001) variables. The new features are: 

1. DataType=0x0001 in EDS is parsed as C++ bool, regarding positive integer values of DefaultValue as "true"
2. It's published as ROS std_msgs/Bool object. 

Addresses https://github.com/ros-industrial/ros_canopen/issues/268

### DIFF
--- a/canopen_chain_node/src/ros_chain.cpp
+++ b/canopen_chain_node/src/ros_chain.cpp
@@ -2,6 +2,7 @@
 
 #include <canopen_chain_node/ros_chain.h>
 
+#include <std_msgs/Bool.h>
 #include <std_msgs/Int8.h>
 #include <std_msgs/Int16.h>
 #include <std_msgs/Int32.h>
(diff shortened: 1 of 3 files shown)


Roboprotos:
Hi,
Thanks for proposing this PR, this feature would be really helpful. 
Since it has passed all the checks, would it be possible to have it merged with the melodic-devel branch ?

mathias-luedtke:
This fix look okay, but I am unsure about PDO support. `canopen_master` does not support bit-wise access (granularity 1).
Unfortunately, I was never able to test this, because none of our devices support boolean objects.

Roboprotos:
I see, I was mainly interested in the DCF parser, to be able to parse boolean datatype from some manufacturer specific objects that can't be mapped to PDO. 

I agree PDO support should be checked before merging, or if canopen_master doesn't support granularity 1, do you think the parser could just reject files which have a mappable boolean object ?

mathias-luedtke:
> do you think the parser could just reject files which have a mappable boolean object ?

I think it should be possible to fail the init service, if a boolean gets mapped.

lkm1321:
Hi guys, thanks for the interest. I do have a device with Boolean variables. I checked that parsing and SDOs are okay, but not PDOs. I'm happy to test whether PDO works. Would that help with having this merged?

Roboprotos:
Hi @lkm1321, were you able to test your PR with PDOs ?

lkm1321:
Hi @Roboprotos, I did a quick test of this feature with our device (Roboteq RGBL1896) but doesn't seem to work correctly. SDO just seems to read 0 even though the underlying value should be 1. I'm not sure if the fault is in `ros_canopen` or in the device's firmware, and I'd need to look at `cansniffer` output to debug, but I'll be away from lab until February so it'll be some time before I do any debugging. I'll update if any developments before that.

PR #420, opened 2021-01-15
Added optional configuration to specify received_messages topic name.

### LINKED ISSUE #419: Missing configuration of topic names in socketcan_bridge
Hello all, 
I've noticed that the socketcan_to_topic bridge doesn't support setting the name of the topic the data is exposed on. I've already implemented a fix for that, and will create a pull request for the fix.

Reasoning behind that change is that we can guarantee that this node can be fitted into a space, which have a specific naming scheme. 

Done:
- socketcan_to_topic_node

Still open: 
- topic_to_socketcan_node

DatDucati:
Would fix #419

### DIFF
--- a/socketcan_bridge/src/socketcan_to_topic.cpp
+++ b/socketcan_bridge/src/socketcan_to_topic.cpp
@@ -52,7 +52,9 @@ namespace socketcan_bridge
   SocketCANToTopic::SocketCANToTopic(ros::NodeHandle* nh, ros::NodeHandle* nh_param,
       can::DriverInterfaceSharedPtr driver)
     {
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Wouldn't this be easier with remapping? Using parameter names for topic names is a bit of an anti-pattern.

DatDucati:
Well, you learn something new every day I guess :) It's up to you if you want to accept this PR, or not, meanwhile I'll try the remap option.

Edit: It works via remap, just like my code does. I'll close this as it's not needed!

mathias-luedtke:
> It's up to you if you want to accept this PR

As @gavanderhoorn already explained: remapping is the preferred way, so there is no need for such a patch.

PR #423, opened 2021-03-01
Refine fatal error configuration.

windelbouwman:
This refines selection of fatal en non-fatal errors.

There was no way to select `BUSOFF` as non fatal. My use case would be to select `BUSOFF` as non fatal and configure the can device with a `restart-ms`.

### DIFF
--- a/socketcan_interface/include/socketcan_interface/socketcan.h
+++ b/socketcan_interface/include/socketcan_interface/socketcan.h
@@ -77,8 +77,18 @@ public:
                                            | CAN_ERR_TRX          /* transceiver status  / data[4]    */
                                            | CAN_ERR_ACK          /* received no ACK on transmission */
                                            );
-      can_err_mask_t fatal_error_mask = parse_error_mask(settings, "fatal_error_mask", fatal_errors) | CAN_ERR_BUSOFF;
-      can_err_mask_t error_mask = parse_error_mask(settings, "error_mask", report_errors | fatal_error_mask) | fatal_error_mask;
+      can_err_mask_t fatal_error_mask = parse_error_mask(settings, "fatal_error_mask", fatal_errors);
+      can_err_mask_t error_mask = parse_error_mask(settings, "error_mask", report_errors | fatal_errors) | fatal_error_mask;
+      if(fatal_error_mask){
+        std::string fatal_errors;
+        translateError(fatal_error_mask, fatal_errors);
+        ROSCANOPEN_INFO("socketcan_interface", "Treating these errors as fatal: " << fatal_errors);
+      }
+      if(error_mask){
+        std::string errors;
+        translateError(error_mask, errors);
+        ROSCANOPEN_INFO("socketcan_interface", "Logging these errors: " << errors);
+      }
       return init(device, loopback, error_mask, fatal_error_mask);
     }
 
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
> There was no way to select BUSOFF as non fatal.

It was written this way on purpose ;)

> My use case would be to select BUSOFF as non fatal and configure the can device with a restart-ms

Does this really work with this fix?
(I did not have a chance to test it)

socketcan_interface/include/socketcan_interface/socketcan.h
```diff
@@ -244,10 +254,13 @@ class SocketCANInterface : public AsioDriver<boost::asio::posix::stream_descript
                 input_.id = frame_.can_id & CAN_EFF_MASK;
                 input_.is_error = 1;
 
-                if (frame_.can_id & fatal_error_mask_) {
+                if (frame_.can_id & error_mask_) {
```
mathias-luedtke, line 257: Not sure about this change..
This triggers the state callback on every error message and might a little bit noisy in general

windelbouwman: My idea was that `error_mask_` is used to test for errors which must be logged. The `fatal_error_mask_` is intended for errors which must put the driver in not ready mode (using `setNotReady`). Is this correct?

I like this idea, apart from the recovery strategy, users can configure which errors should be fatal. Another idea would be to simply log all errors, en only allow to configure which errors are fatal.

The concept of "fatal" being the behavior that the driver is put into not ready mode.

windelbouwman: > This triggers the state callback on every error message and might a little bit noisy in general

Error's should only happen every now and then (arbitration error, ack error). If there are many error's on the bus, it's probably okay to log them? Another strategy could be to rate limit the amount of errors logged, and summarize them, but this would be more work.

windelbouwman: Btw, the reason why this change is important, is that in case of an error, every now and then, the driver hangs on "failed to send message".

Now that I think about it, in the current implementation all errors, except the BUS_OFF error can be configured to be ignored, right? Setting the whole bunch of parameters to false will do the trick!

windelbouwman:
> It was written this way on purpose ;)

Why would the BUS_OFF error be a special case? I think it's fair to treat it as the other errors. What will happen with BUS_OFF, is that the output queue will fill up, and eventually the send asio call will fail / block.

socketcan_interface/include/socketcan_interface/socketcan.h
mathias-luedtke: > My idea was that error_mask_ is used to test for errors which must be logged.

`error_mask` entry will be reported to the error frames callback!

We could add a `logging_mask` for this purpose, which could default to `error_mask`.

mathias-luedtke:
> What will happen with BUS_OFF, is that the output queue will fill up, and eventually the send asio call will fail / block.

In our usecase, the output queue will fill up immediately and then the kernel driver would close the socket..

> Why would the BUS_OFF error be a special case? I think it's fair to treat it as the other errors

Agreed.
Again, I did not test this special case.

socketcan_interface/include/socketcan_interface/socketcan.h
windelbouwman: I'm not in favor of adding extra configration. I suggest to move the logging inside the fatal error clause?

I guess the error frame callback is call by calling `setInternalError`?

PR #438, opened 2021-07-06
Setup CANopen nodes in listed order if applicable

mathias-luedtke:


mathias-luedtke:
Tested on hardware, works as expected now.

PR #442, opened 2021-08-19
add configurable homing timeout

sgermanserrano:
Enables to have a configurable homing timeout for slower joints, e.g.

```
defaults: # optional, all defaults can be overwritten per node
  eds_pkg: my_config_package # optional package  name for relative path
  eds_file: "my_config.dcf" # path to EDS/DCF file
  motor_layer:
    homing_timeout: 60
    "1016sub1" : "0x7F0064" # heartbeat timeout of 100 ms for master at 127
    "1017": "100" # heartbeat producer
```
It defaults to 10s to keep the original behaviour

Signed-off-by: Servando German Serrano <servando.german.serrano@gmail.com>

### DIFF
--- a/canopen_402/src/motor.cpp
+++ b/canopen_402/src/motor.cpp
@@ -242,6 +242,61 @@ bool DefaultHomingMode::executeHoming(canopen::LayerStatus &status) {
     return error(status, "something went wrong while homing");
 }
 
+bool DefaultHomingMode::executeHoming(canopen::LayerStatus &status, const boost::chrono::seconds &homing_timeout) {
+    if(!homing_method_.valid()){
+        return error(status, "homing method entry is not valid");
+    }
+
+    if(homing_method_.get_cached() == 0){
+        return true;
+    }
+
+    time_point prepare_time = get_abs_time(boost::chrono::seconds(1));
+    // ensure homing is not running
+    boost::mutex::scoped_lock lock(mutex_);
+    if(!cond_.wait_until(lock, prepare_time, masked_status_not_equal<MASK_Error | MASK_Reached, 0> (status_))){
+        return error(status, "could not prepare homing");
+    }
+    if(status_ & MASK_Error){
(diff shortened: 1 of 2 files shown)


canopen_402/src/motor.cpp
```diff
@@ -504,6 +508,8 @@ void Motor402::handleInit(LayerStatus &status){
         return;
     }
 
+    homing->setHomingTimeout(homing_timeout_);
```
mathias-luedtke: Why did you create a new function and did not  pass it as a parameter to `executeHoming`?

mathias-luedtke (commented):
Thanks! I had a similar patch in mind. Right now I am just not sure, if we should break the ABI like this.

canopen_402/src/motor.cpp
sgermanserrano: @ipa-mdl tbh, I didn't know which option was better, I'll refactor to have it as:
`executeHoming(canopen::LayerStatus &status, const boost::chrono::seconds &homing_timeout)`
and pass it with the call to `executeHoming`

mathias-luedtke: Me neither ;)

That's why I wanted to know why you chose that option.

> I'll refactor to have it as:
executeHoming(canopen::LayerStatus &status, const boost::chrono::seconds &homing_timeout)
and pass it with the call to executeHoming

:+1:

Please keep the function without the argument as well (stable API) and call the new one with a timeout of 10s (as before).
I will try to figure out, if we can avoid the ABI break in `Motor402`.

sgermanserrano: I've added the new function and replace the call to `executeHoming` with a default timeout of 10s

ivan140:
Any updates or help needed on this pull request? I've tested it on a super slow joint and the homing was working really good, so looking forward for a merge :)

amilcarlucas:
Ping. Please merge

Issue #467, opened 2022-09-01
Recovery from USB disconnection


BenArtes:
We're using socketcan_interface for communicating with CAN devices. Due to some electrical issues power / USB connections have dropped out and communications fail. Obviously we're investigating the hardware issues causing this, but I attempted to add automated software recovery.

* Made sure that Linux was configured for hotplug functionality and automatically configuring / bringing up the CAN interfaces
* Tried to driver->recover

The recover call seems to hang. I'm investigating the cause, but is there a suggested way of handling disconnection / re-connection of a CAN interface?

BenArtes:
Just got back to working on this.

The direct cause of the lockup I was experiencing was that I was calling driver_->recover() from within the stateCallback. This locked up the thread while it waited for itself to release the state mutex.

After the driver would 'recover', but it wouldn't receive any new messages. I believe this is because I was using the ThreadedSocketCANInterface which was missing an overridden 'recover' method to restart the run_thread. I applied the modifications in #469 which resolved my issues.

PR #469, opened 2022-09-01
Added can device recovery code

akashmod:
This is to solve issue 468. A ros timer is started to check the driver state after every specified duration and to reinitialize the driver for the node if the driver has recovered.

### DIFF
--- a/socketcan_bridge/include/socketcan_bridge/socketcan_to_topic.h
+++ b/socketcan_bridge/include/socketcan_bridge/socketcan_to_topic.h
@@ -50,10 +50,11 @@ class SocketCANToTopic
 
     can::FrameListenerConstSharedPtr frame_listener_;
     can::StateListenerConstSharedPtr state_listener_;
-
+    ros::Timer timer;
 
     void frameCallback(const can::Frame& f);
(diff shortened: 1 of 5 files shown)


BenArtes:
Summary notes for any reviewers: This covers similar ground as PR #249 (5 years old and waiting for changes) and #310 (4 years old and VERY minimal).

I think whatever happens to the socketcan_bridge auto-recovery code, the ThreadedSocketCANInterface should be merged as a (potentially) separate PR as it fixes a seemingly missing piece from ThreadedSocketCANInterface recovery code (as I ran into in issue #467).

BenArtes:
Also, shouldn't ThreadedSocketCANInterface `virtual void recover` have a bool return type?

akashmod:
The virtual void recover should have a bool return type yes. That needs to be corrected.

carlesu:
Hello, Tank you very much!! This was giving me a big headache, because we actually have a device that has on/off using dead man button and had to restart the node in order to send new data. Been following #467 since then, this seems to solve the problem.

Is it possible to know approximately when could this be updated to noetic release?

Thank you again for the effort!

PR #471, opened 2022-09-20
Fix edge case in $NODEID parsing within EDS object dictionary values

rugged-robotics:
This fixes the case where if an EDS contains a value of `literal+$NODEID` it is currently always interpreted as the literal value. This results in an odd failure mode when object 0x1200 is defined as `0x600+$NODEID` where the canopen master will send SDO commands to the CAN ID `0x600`.

Fix added as well as relevant test cases.


mathias-luedtke (approved):
Thank you very much for your contribution!
@ipa-hsd: please review, merge and release.

Issue #468, opened 2022-09-01
Can device goes into an error state and the socketcan node never recovers


akashmod:
We have noticed that the can communication drops off sometimes because the socketcan node goes into an error state, if the driver goes into an error state and never recovers even if the driver is back online. Respawning the node solves the issue, which makes us believe that the issue is in the node itself.

mathias-luedtke:
If you are talking about the socketcan_bridge node, then this is a duplicate of https://github.com/ros-industrial/ros_canopen/issues/244.
The other nodes have a `recover` service..

PR #474, opened 2022-10-29
To resolve the socketcan interface dying due to can errors and failing to recover

akashmod:
This is to solve issue #468. A ros timer is started to check the driver state after every specified duration and to reinitialize the driver for the node if the driver has recovered.

### DIFF
--- a/socketcan_bridge/src/socketcan_to_topic.cpp
+++ b/socketcan_bridge/src/socketcan_to_topic.cpp
@@ -55,6 +55,7 @@ namespace socketcan_bridge
       can_topic_ = nh->advertise<can_msgs::Frame>("received_messages",
                                                   nh_param->param("received_messages_queue_size", 10));
       driver_ = driver;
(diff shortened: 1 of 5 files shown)


socketcan_bridge/src/socketcan_to_topic.cpp
```diff
@@ -55,6 +55,7 @@ namespace socketcan_bridge
       can_topic_ = nh->advertise<can_msgs::Frame>("received_messages",
                                                   nh_param->param("received_messages_queue_size", 10));
       driver_ = driver;
+      timer = nh->createTimer(ros::Duration(0.1), std::bind(&SocketCANToTopic::timerCallback, this, std::placeholders::_1));
```
hellantos, line 58: This is a polling approach. We should at least be able to configure the duration via for example a ROS parameter. Also there should be an option to disable this behaviour.

Timple: Perhaps by treating a zero duration as disabled. Zero is a common exception used for this and save a new parameter.

PR #475, opened 2023-04-11
Use correct index with check_com_changed

s-hall:
Now correctly checks if the communication parameters need to be written. I guess this issue was masked because the mapping parameters are usually changed as well. Came across this when configuring a drive (AMC DZCANTE-012L080) with read-only RPDO mapping parameters and read-write communication parameters.


mathias-luedtke (approved):

mathias-luedtke:
I think the fix is correct, but I cannot test it :/

PR #480, opened 2023-08-07
fix: error: 'set' in namespace 'std' does not name a template type

muellerbernd:


mathias-luedtke (approved):
Makes sense!
Thanks :)

PR #483, opened 2024-01-29
use default C++ standard instead of 14

lucasw:
I get this when trying to compile in Ubuntu 22.04:

```
.../src/drivers/ros_canopen/socketcan_bridge/src/socketcan_to_topic.cpp:28:
/usr/include/log4cxx/boost-std-configuration.h:10:18: error: ‘shared_mutex’ in namespace ‘std’ does not name a type
    10 |     typedef std::shared_mutex shared_mutex;
       |                  ^~~~~~~~~~~~
 /usr/include/log4cxx/boost-std-configuration.h:10:13: note: ‘std::shared_mutex’ is only available from C++17 onwards
    10 |     typedef std::shared_mutex shared_mutex;
       |             ^~~
make[2]: *** [CMakeFiles/socketcan_to_topic.dir/build.make:90: CMakeFiles/socketcan_to_topic.dir/src/socketcan_to_topic.cpp.o] 
```

Using the default 22.04 C++ standard (17 I think) fixes it.

This would more accurately be part of a `noetic` branch which would mean the default C++ standard is already 14.

### DIFF
--- a/can_msgs/CMakeLists.txt
+++ b/can_msgs/CMakeLists.txt
@@ -1,11 +1,6 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(can_msgs)
 
-if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD_REQUIRED ON)
-  set(CMAKE_CXX_STANDARD 14)
-endif()
-
 find_package(catkin REQUIRED
   COMPONENTS
     message_generation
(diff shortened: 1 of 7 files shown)


mathias-luedtke (changes requested):
Please rebase (#480 got merged) and while you are at it better fix the typo ;)

mathias-luedtke:
https://www.ros.org/reps/rep-0003.html#c already defines C++14 as the default for melodic

mathias-luedtke (approved):

